PR #30, opened 2015-02-18
Get the species trees from the compara db

muffato:
Summary of changes:
 - there can be more than 2 trees. The Perl code will fetch all the trees that the compara db is exposing
 - species filters and the default tree are defined in the compara db


muffato:
My sandbox: http://enssand-01.internal.sanger.ac.uk:9073/info/about/speciestree.html

ens-ma7:
Hi Matthieu
Thats quite a lot of changes, need to have a good look at them. I am a bit busy with trying to get something else out. once i am done with it i will look at this.

Thanks
Ridwan

PR #40, opened 2015-08-05
Tree shrew was not counted as a placental mammal. Now in the category "Rodents etc"

muffato:
Fixes part of RT#482574


muffato:
I'm not sure exactly which ones are needed but on my sandbox I deleted web_tree.packed and spp_tree.packed and the change was then picked up at the server start

PR #43, opened 2015-10-07
Don't die if examples are not available

nicklangridge:
Previously if examples are not present there was a runtime error: hash- or arrayref expected (not a simple scalar, use allow_nonref to allow this) at /nfs/public/rw/ensembl/libs/perl/share/perl5/JSON.pm line 154


willmclaren:
Looks fine to me (most of this code is mine originally I think) but I'll let web merge as it's their repo.

PR #48, opened 2016-02-08
[CAFE view] Use the precomputed JSON

muffato:
cf https://github.com/Ensembl/ensembl-webcode/pull/372

### DIFF
--- a/widgets/modules/EnsEMBL/Web/Component/Gene/SpeciesTree.pm
+++ b/widgets/modules/EnsEMBL/Web/Component/Gene/SpeciesTree.pm
@@ -28,6 +28,23 @@ use JSON qw(to_json);
 
 use previous qw(content);
 
+sub get_json {
+  my $self   = shift;
+  my $cdb    = shift;
+
+  my $object = shift || $self->object || $self->hub->core_object('gene');
+  my $member = $object->get_compara_Member($cdb);
+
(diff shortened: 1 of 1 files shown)


ens-ma7:
Hi Matthieu
The sub get_details is going to overwrite the one from the main code (ensembl-webcode/modules/EnsEMBL/Web/Component/Gene/SpeciesTree.pm). Are you sure you want to do this? 
If so, i will suggest to move the sub get_details to main code and need to make sure it will work with the fallback view.

muffato:
That's actually why I'm doing it this way: the precomputed JSON is only useful for the new view since the fallback view is Perl only. So I think it's fine to override get_details in public-plugins/widget. Both views have their own `get_details()` and `content()`

ens-ma7:
Sorry Matthieu, doing it this way will make it use the get_detail inside the widget plugin, this is how the plugins work. Any function inside a plugin will overwrite the main one. 
The solution to this is to have a different get_detail function in the plugin (say get_json) and then call get-json instead of get_details inside the plugins component code.
Hope that makes sense, let me know if there is anything not clear.

muffato:
Aaaah I understand: the widget modules are all in PERL5LIB of the server, regardless of the whether the client needs the fallback view or the JavaScript one. Got it
I will amend the commit and update the PR

ens-ma7:
Thank you

muffato:
Done. It still works on my sandbox

PR #50, opened 2016-02-25
update to VEP docs reflecting move to Bio::DB::HTS module from faidx_xs

rishidev:
Changes have been made to the VEP to use Faidx (FASTA indexing from HTSlib) from Bio::DB::HTS rather than its own module, as detailed in the merged pull request at https://github.com/Ensembl/ensembl-tools/pull/12. These documentation changes accompany this.


willmclaren:
Looks fine, though whatever editor you are using seems to insist on changing every bit of whitespace in the file...

PR #56, opened 2016-04-15
Genebuild help pages: updated links and titles.

ens-carlos:


ens-ap5:
Hi Carlos

This is mostly fine, but the title tag contents are too long - we have to keep them brief because they're used in the lefthand side navigation. Please could you amend them to be as short as possible, e.g. 'Ensembl Annotation' instead of 'The Ensembl Annotation Process'? You can keep the full text for the h1 tag, which is displayed within the page.

Thanks

Anne

ens-carlos:
Hi Anne,

Sorry that I didn't think about that. Some titles were too long as you pointed out. I've shortened them and the h1 too as we were aiming to make them match. I hope it looks better now.

Cheers,
Carlos

ens-ap5:
No worries! As I said, the <h1> tags can contain longer text if you want to be more precise, it's just the <title> tags that we like to be short.

PR #58, opened 2016-05-10
Need to quote the option value in find expression

brucebolt:
It may contain characters like commas or spaces.  Similar to b7a4335b32798933dfe872f7a7cfdeee1cd0e0e0


ens-ap5:
Could you submit this on the postreleasefix/84 branch, please? That makes it easier for us to merge into master as well as the release. Ta!

brucebolt:
Submitted on postreleasefix/84 branch.

PR #71, opened 2016-10-24
Point to the right URL for track hub docs

andrewyatz:
Highlighted by Ann Zweig we were pointing to the wrong docs.


ens-hsr:
Merged to master too: 87544b1e

PR #77, opened 2017-02-27
Modifications to Variation API tutorial

helensch:
Changes to replace deprecated methods that will be removed in release 88.
Change examples to variants that return results
@ens-lgil Please can you review.


ens-lgil:
Look ok @helensch!

PR #81, opened 2017-05-02
Changes made to tutorials page

ens-emily:


ens-ds23:
It looks like this is the same PR as #82 but against the wrong branch, so I can close this without merging?

PR #84, opened 2017-07-26
Switching to original apache 2 license text

andrewyatz:
**Please wait before accepting in**

Switching back to the original apache license text but also added a new section for Ensembl.


ens-hsr:
un-merged again

PR #88, opened 2017-07-28
Updates to rodents static content

leannehaggerty:
Ok, this should be right, the files have been renamed to start with a capital letter and everything should be beautiful... let me know if it isn't

Thanks!


ens-ap5:
There are a few files at the bottom of the list that are still lower case, and I think the chinese hamster needs the cell line in its name. Have a look on staging to see each species' URL.

leannehaggerty:
Are you sure those aren’t just deleted files (in red)? e.g. ensembl/htdocs/ssi/species/ictidomys_tridecemlineatus_assembly.html <https://github.com/Ensembl/public-plugins/pull/88/files#diff-139fdd0c451db2d2a2e04d88543c2abb>has been deleted and a new file has taken it’s place ensembl/htdocs/ssi/species/Ictidomys_tridecemlineatus_assembly.html <https://github.com/Ensembl/public-plugins/pull/88/files#diff-ab442a00b77dd7798b2142cc157a1fac>

ens-ap5:
Ah yes, you're right - sorry, it's Friday afternoon and our sandboxes are very sluggish, so we're all tired and frustrated in this office!

leannehaggerty:
Completely understandable, we have a very similar feeling in this office, we’ve almost made it to the weekend though!

Thanks for all your help :)

PR #97, opened 2017-11-07
beta url retired for trackhub registry

magaliruffier:
the beta url does not exist any more, updated the documentation to link to the public URL


docs/htdocs/info/website/public_trackhubs.html
```diff
@@ -10,7 +10,7 @@
 
 <p>Ensembl provides support for track hubs as a way of sharing large quantities of annotation
 over the internet. A list of public track hubs is available at the <a
-        href="http://beta.trackhubregistry.org/">Trackhub Registry</a> based at
+        href="http://trackhubregistry.org/">Trackhub Registry</a> based at
```
avullo: I would put https here, otherwise some browsers (e.g. chrome) will show problems.

PR #106, opened 2017-11-23
Added SIFT and Polyphen threshold tables

aparton:
Due to lack of sandbox, I cannot easily verify that the styling of the tables is ok, can someone take a look?


willmclaren:
The page looks fine, but you'll need to add this as a commit against master rather than release/90 and re-open the pull request

PR #113, opened 2018-01-09
update VEP command

at7:
use ./vep instead of vep.pl

### DIFF
--- a/tools/modules/EnsEMBL/Web/Component/Tools/VEP/TicketDetails.pm
+++ b/tools/modules/EnsEMBL/Web/Component/Tools/VEP/TicketDetails.pm
@@ -120,7 +120,7 @@ sub job_details_table {
   );
 
   ## create command line that users can cut and paste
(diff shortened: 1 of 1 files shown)


tools/modules/EnsEMBL/Web/Component/Tools/VEP/TicketDetails.pm
```diff
@@ -120,7 +120,7 @@ sub job_details_table {
   );
 
   ## create command line that users can cut and paste
-  my $command_string = 'perl vep.pl';
+  my $command_string = 'perl ./vep';
```
ens-lgil: Should it be my `$command_string = './vep';` instead of `my $command_string = 'perl ./vep';` (as Helen reported) to match the VEP doc ?
i.e. http://www.ensembl.org/info/docs/tools/vep/script/vep_options.html

at7:
yes

PR #123, opened 2018-02-14
remove redundant header

ima23:


ens-ap5:
Hi Irina, do we know if either of those IDs are used anywhere to jump directly to that section, e.g. with #sv on the end of the URL?

ima23:
@ens-ap5 Looked with grep and did not find anything.

ens-ap5:
OK - thanks for checking!

PR #134, opened 2018-04-11
RNA Sequence Support for BLAST

imran-ebi:
This pull request is for ENSWEB-4464. 'U' has been added to check DNA sequence type. The 'U' is replaced by 'T' to prevent BLAST jobs from failing. Flags are used to display a message about this to the user. The sandbox links is: http://ves-hx-78.ebi.ac.uk:8310/Multi/Tools/Blast.


imran-ebi:
I have pushed the fix for RT #264366 as well.

imran-ebi:
I have changed the REST URL to HTTPS protocol as it does not cause any problems when the site is using HTTP.

PR #145, opened 2018-07-18
update cat example

sarahhunt:
Previous example no longer available


ens-ma7:
@sarahhunt  is that for VEP only?

sarahhunt:
It is to update the example phenotype here: http://www.ensembl.org/Felis_catus/Info/Index

ens-ma7:
Ok thanks .... I will push it to the live servers now.

PR #149, opened 2018-09-06
Update the VEP result page by hidding the empty columns in order to r…

ens-lgil:
…educe the number of columns visible on the wide table


ens-st3:
Can you take a look @ens-ma7 ?

ens-ma7:
Hi @ens-lgil can i have your sandbox URL please (with a result example)

Thanks
Ridwan

ens-ma7:
where margin-right is used in the inline styling, can you please use the class right-margin instead

PR #152, opened 2018-09-20
Update Core schema documentation to release 94

mkszuba:
The title says it all.


ens-ap5:
In future, could you make your pull request against e.g. postreleasefix/94 - I realise the timing was probably not favourable on this occasion, since we only branched yesterday and the postreleasefix/94 branch would not have been available right away.

PR #178, opened 2019-01-09
resolve cell type formatting conflict

at7:
This resolves a conflict between the two ways cell type names have been generated in the public-plugins code and ensembl-vep code. Reported by Steve in this [ticket](https://helpdesk.ebi.ac.uk/Ticket/Display.html?id=317126) which I cannot see. And working with the fix on my [sandbox](http://ves-hx2-76.ebi.ac.uk:5040/Homo_sapiens/Tools/VEP/Results?db=core;tl=gZrPUyEH29SboPVa-256). It is broken on 94, 95. Let me know to which other branches you want the PR.


ens-st3:
Thanks Anja - please can you resubmit this against postreleasefix/95 ?

PR #184, opened 2019-02-15
updates to describe species supported direct from EVA

sarahhunt:
Also moved broad data descriptions up a level and the information on viewing tracks from spectifi data sources to the sources page.


ens-lgil:
Some of your changes are (re)generated from our documentation scripts.
I will have a look and updates them regarding your changes

sarahhunt:
Thanks @ens-lgil - I thought that may be the case.

PR #193, opened 2019-03-08
add example variant

at7:
The [example variant](http://ves-hx2-76.ebi.ac.uk:5040/Chlorocebus_sabaeus/Info/Index) is currently not working. Vervet doesn't have any variant ids in the database which can be used for search. We need to add the location with vf= to the url in order to navigate to the variant summary page. Is there a way of encoding this in the ini file without ending up with a mangled up [URL](http://ves-hx2-76.ebi.ac.uk:5040/Chlorocebus_sabaeus/Variation/Explore?v=%26vf%3D11:6825247:G_A:PRJEB22989) following the link from the [Example variant](http://ves-hx2-76.ebi.ac.uk:5040/Chlorocebus_sabaeus/Info/Index)? Thanks.


ens-ap5:
I think it will need a tweak in the main ensembl-webcode to get rid of the ampersand. Could you submit a JIRA ticket?

at7:
Thank you Anne. The ticket is [here](https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-4939).

PR #194, opened 2019-03-11
use release version

at7:
I updated the path to use the release version. This will be important to get the correct data when checking archive sites.

### DIFF
--- a/ensembl/conf/json/bos_taurus_vcf.json
+++ b/ensembl/conf/json/bos_taurus_vcf.json
@@ -17,7 +17,7 @@
       "species": "bos_taurus",
       "assembly": "ARS-UCD1.2",
       "type": "remote",
-      "filename_template" : "ftp://ftp.ensembl.org/pub/release-96/compara/conservation_scores/88_mammals.gerp_conservation_score/gerp_conservation_scores.bos_taurus.ARS-UCD1.2.bw",
+      "filename_template" : "/88_mammals.gerp_conservation_score/gerp_conservation_scores.bos_taurus.ARS-UCD1.2.bw",
       "annotation_type": "gerp"
     }
   ]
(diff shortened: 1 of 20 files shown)


ens-ma7:
@at7 but that means for each release we will have to update the path?

at7:
The data is not static and compara includes new species every release. If we keep updating the release version and the new new name of the alignment directory (from 75_mammals.gerp_conservation_score in release-95 to 88_mammals.gerp_conservation_score in release-96) we will be backwards compatible for archive sites. They might not always increase but I think this way we are safe?

ens-ap5:
Given that the first part of the URL is identical in each species ('ftp://ftp.ensembl.org/pub/current_compara/conservation_scores/'), it would make sense to generate it in the code that consumes the JSON, instead of having the full path hard-coded. Then we wouldn't need to update it every release, just when the number of mammals in the GERP alignment changes.

at7:
I like that. Shall we create a ticket for this for next time or is there time do it for 96? I can adjust the paths accordingly in the config files.

ens-ap5:
It should be a simple change. Put in a ticket now and I'll pick it up, like we did for the vervet variation link.

at7:
Here is the ticket: https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-4943
Thank you

ens-ma7:
Thanks @ens-ap5, i was going to suggest the same thing, you are too quick in responding, i was confused, i thought i was replying on an all PR.

I will leave this PR open until you have your changes in @ens-ap5 , feel free to close it once you are happy.

ens-ap5:
Sorry, I thought I'd chime in since I saw an obvious solution :)

ens-ma7:
No problem at all, its just that within 5 minutes there was like 4 comments on the PR :) anyway its all yours now :)

at7:
Is this what you had in mind? If yes, I will update the other species as well.

ens-ap5:
Yes, that's perfect!

PR #199, opened 2019-03-28
Update PDBe JS regarding some data changes in Ensembl REST

ens-lgil:
Fixing a missing Core data change I couldn't test before the Test REST server was available.
See an example with the fix on my sandbox:
http://ves-hx2-76.ebi.ac.uk:5060/Homo_sapiens/Variation/PDB?v=rs699


ens-ma7:
All good, i have merged the changes and push to the test checkout and restarted test, you should be able to test this on test.ensembl.org

ens-lgil:
Thank you Ridwan!

PR #202, opened 2019-04-02
Updates to 'Command Line Equivalent' for web VEP

aparton:
For https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-1475


aparton:
I assumed I was too late for 96, but if you're happy to merge then I'm happy to put it on there. I've created a new PR

PR #204, opened 2019-04-09
fix add cadd files parameter grch38

ima23:
Added missing parameter for CADD files for e96 GRCh38 web.

Example of sandbox with results showing for CADD, columns CADD PHRED, CADD RAW:
http://ves-hx2-76.ebi.ac.uk:6080/Homo_sapiens/Tools/VEP/Results?db=core;tl=SboxrAwK3AT2CPEt-655


ima23:
@jyothishnt : I've updated now the file with the correct file name. Thanks.

PR #206, opened 2019-04-26
Hide the 'Phenotypes data' form for the species without phenotype data in the Web VEP input form

ens-lgil:


ens-lgil:
Here is the example on my sandbox:
http://ves-hx2-76.ebi.ac.uk:5060/Multi/Tools/VEP?db=core

The `Phenotype data` form is in the `Additional annotations` fieldset

PR #209, opened 2019-05-10
Do not show quick links when backend tells not to

azangru:
JIRA ticket: https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-3464

### Context
The original task in the JIRA ticket was not to show a quick link to orthologues in a search result if there are none. Solr now returns a value to indicate whether there are any orthologues. Here's what this looks like in the api response (see [example request](http://www.ensembl.org/Multi/Ajax/search?q=(+NOT+species%3Axxx+)+AND+(+ENSG00000262826+)+AND+(+NOT+species%3Ayyy+)&fq=&rows=1&facet.field=species&facet.field=feature_type&facet.field=strain&facet.mincount=1&facet=true)):

image

There surfaced, however, another problem. Sometimes a search result will have no valid quick links associated with it (for example, see the last search match on [this page](http://www.ensembl.org/Multi/Search/Results?q=ENSG00000262826;page=1)). Currently, when this is the case solr conveniently puts the string `"none"` in the `quick_links` array (see [sample request](http://www.ensembl.org/Multi/Ajax/search?q=(+ENSG00000262826%5E316+AND+species%3A%22CrossSpecies%22+)+OR+(+ENSG00000262826%5E190+AND+species%3A%22Human%22+)+OR+(+ENSG00000262826%5E80+AND+species%3A%22Mouse%22+)+OR+(+ENSG00000262826+AND+species%3A%22Zebrafish%22+)&fq=(++(++species%3A%22CrossSpecies%22+AND+(+reference_strain%3A1+)++)++OR++(++species%3A%22Human%22+AND+(+reference_strain%3A1+)++)++OR++(++species%3A%22Mouse%22+AND+(+reference_strain%3A1+)++)++OR++(++species%3A%22Zebrafish%22+AND+(+reference_strain%3A1+)++)++)&hl=true&hl.fl=_hr&hl.fl=content&hl.fl=description&hl.fragsize=500&start=0&rows=10)):

image

So what we _could_ do is also to check for this string when rendering the list of quick links (which is being done in this PR). Although @ens-ds23 [thinks](https://genomes-ebi.slack.com/archives/C0MUF0C3Y/p1557408237327900) that this is incorrect behaviour of solr or of the api:

> I don't think that `none` should be returned by the server. At least, I don't think anyone anticipated it. Probably best to chase it into the SOLR result proper: it is probably a mistake at build time. @stevet is usually quite keen we push stuff back into build time (sensibly), but there's a big infrastructure for hacks that haven't made it or is too complex to do, still used in the client side (so we *can* but probably *shouldn't*). Probably best to create a JIRA for it so that it doesn't get lost.

### Possible complications

I do not know what values inside the `quick_links` array are possible. The current solution expects that it will contain strings of format `"<link_title>: number"` or `"none"`.


solr/src/search.coffee
```diff
@@ -738,6 +741,22 @@ body_quicklinks = () ->
...
 
+should_exclude_from_quicklinks = (title, doc) ->
+  # if a doc contains a quick_links field (which is an array)
+  # and if that array contains a string of a format "title: 0"
+  # (where title equals the first argument of the function,
+  # this quick link should not be shown on the page
+  rules = doc.quick_links || []
+
+  for rule in rules
+    if rule == 'none'
```
azangru: this relies on the current (possibly incorrect) behaviour of solr (see PR description for details)

azangru:
@ens-st3 @ens-ds23 could you also leave your thoughts here?

ens-ds23:
Fine by me.

PR #244, opened 2019-08-30
e98_pdfs_htmls

ndliberial:
Html and genebuild pdf files for e98 fish, xenopus tropicalis, pigs and larimichthys crocea.
Pigs do not have any html files.


ens-ap5:
I'm not sure what's gone wrong here, but there are no file changes. Please check your branch and recommit.

ens-ap5:
Hi Denye, could you please remove the species descriptions from the annotation.html files? We moved this content to separate files a few releases ago and agreed with Fergal that we would no longer be importing text programmatically from Wikipedia because the length and quality is completely unreliable.

ndliberial:
Hi Anne,
Thibaut is looking into it as we have a script that automatically generates these html files.

ndliberial:
Hi Anne,
Just made the requested changes now

ens-ap5:
Thanks - I'll look at it shortly.

ens-ap5:
OK, I'm going to merge this in, but please be more careful in future when selecting branches. Your initial commits were actually on the staging branch, so they would have been applied to the staging site without a PR being performed - this PR only removes the unwanted species descriptions, which initially confused me. 

In future, please create a feature branch first before committing your changes, and always submit the request against the staging branch. Thanks!

PR #253, opened 2019-09-16
remove default format for custom flag 

ima23:
sandbox: http://ves-hx2-76.ebi.ac.uk:6080/info/docs/tools/vep/script/vep_custom.html


ens-ma7:
Test has also been updated.

PR #240, opened 2019-08-21
Add new job button on tools result page

msmanoj:
Related Ticket:
https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5053

Description:
Added new job button to the following tools result page:

1. Blast - Top & bottom
http://ves-hx2-76.ebi.ac.uk:42229/Homo_sapiens/Tools/Blast/Results?db=core;expand_form=true;tl=cRCulE5JJEN87Ysh-877
2. VEP - Top & bottom
http://ves-hx2-76.ebi.ac.uk:42229/Homo_sapiens/Tools/VEP/Results?db=core;expand_form=true;tl=6VBkpE2WUATICMqN-733
3. LD - Top - Link not available
4. IDM - Top & bottom
http://ves-hx2-76.ebi.ac.uk:42229/Homo_sapiens/Tools/IDMapper/Results?db=core;tl=Di3pUFNr5WUf6IAb-640
5. VCF - Top & bottom
http://ves-hx2-76.ebi.ac.uk:42229/Mus_musculus/Tools/VcftoPed/Results?db=core;expand_form=true;tl=IoqoESIZ48RWtGer-70
6. AF - Top - Link not available

PR #254, opened 2019-09-16
Fix tools data slicer results page

msmanoj:
**Description:**
This PR is to fix a bug with Tools-> Data Slicer -> Results page, introduced in the following PR:
https://github.com/Ensembl/public-plugins/pull/240

**Views affected:**
http://ves-hx2-76.ebi.ac.uk:42228/Homo_sapiens/Tools/DataSlicer/Results?tl=sotkTt3lgIfNX8sw-994


ens-ma7:
Looks good to me. The only thing that we couldnt test was allele frequency which we can test on test grch37

PR #260, opened 2019-09-20
replace variation_set_1kg_3 with gnomad

at7:
We didn't correctly set which variants belong to the set "1000 Genomes 3 - All" which means that the default track variation_set_1kg_3 would be empty. We switched the default track to variation_set_gnomAD instead.


sarahhunt:
@ens-st3 - is there still time to get this one in? Region In Detail will look sadly bereft without it.

ens-st3:
Yes. Thanks for the nudge.

PR #263, opened 2019-09-26
minor changes to static content for Macaca mulatta

osagiei:


ens-ap5:
I'm sorry, but we can't accept the change to the assembly file. It includes the description from Wikipedia, which I previously pointed out was not acceptable, and lacks the link to the genebuild PDF. If your pipeline can't get this right, it would be better to submit the revised text manually.

PR #265, opened 2019-10-01
Make the VEP Docker installation documentation less confusing, to avo…

ens-lgil:
…id having users trying to (re)install the API

Would it be possible to push it to the live website, please ?


ens-lgil:
Merged to `master` and `release/98`

PR #266, opened 2019-10-04
Replace the PDBe URLs by using HTTPS by default

ens-lgil:
This should fix the PDBe viewer issue when using Ensembl with the HTTP protocol.


ens-ma7:
Thanks for this Laurent, its all merged and push to the live server. Probably your last web PR :(

Thanks for all your PR/fix over the time.  👍

ens-lgil:
Thanks @ens-ma7!
Just tested it on the live website and it works!

PR #267, opened 2019-10-14
Fix typo in fields option

helensch:


azangru:
Should this go in the master branch (will go live during next release) or in the postreleasefix/98 branch (will go live before the next release)?

helensch:
It is only for master - to go live during next release

PR #289, opened 2019-11-13
Add tracking to VEP tool run buttons

msmanoj:
Added tracking to the `Run` and `Run Instant VEP` buttons on the VEP tool page.

Sandbox URL:
http://ves-hx2-76.ebi.ac.uk:42228/Homo_sapiens/Tools/VEP?db=core

### DIFF
--- a/ga/htdocs/components/95_ga_configs.js
+++ b/ga/htdocs/components/95_ga_configs.js
@@ -758,6 +758,31 @@ Ensembl.GA.eventConfigs.push(
                       },
     action          : function () { return this.getURL(); },
     label           : function() { return $(this.currentTarget).text(); }
+  },
+
+  // Track Run VEP button
+  {
+    id              : 'RunVEPButton',
+    event           : 'click',
+    wrapper         : '.ajax.initial_panel',
+    selector        : 'input.run_button',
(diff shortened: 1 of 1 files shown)


ga/htdocs/components/95_ga_configs.js
```diff
@@ -758,6 +758,31 @@ Ensembl.GA.eventConfigs.push(
...
+
+  // Track Run VEP button
+  {
+    id              : 'RunVEPButton',
+    event           : 'click',
+    wrapper         : '.ajax.initial_panel',
+    selector        : 'input.run_button',
+    category        : 'RunVEPButton',
+    action          : 'Tools/VEP',
+    label           : ''
```
ens-ma7: i would advise to put something in the labels for both, in google analytics you can view events by label and we occasionally filter by label when viewing reports in google analytics. Otherwise it will be blank for these. Maybe something like 'VEP Instant Button' and 'Run VEP Button' will do.

ens-ma7: Also change category to 'Tools' for both. @ens-st3 what do you think about the name for the category, will 'tools' be good or is too generic?

PR #296, opened 2019-12-05
Adds --custom_multi_allelic to VEP docs

aparton:


at7:
Almost good. The descriptions needs to be a bit clearer.

at7:
@aparton Can you please update the PR to your latest suggestion. Thank you

> By default, comma separated lists found within the INFO field of custom annotation VCFs are assumed to be allele specific. For example, a variant with allele_string A/G/C with associated custom annotation 'single,double,triple' will associate triple with C, double with G and single with A. This flag instructs VEP to return all annotations for all alleles.

PR #298, opened 2019-12-09
Fix link for embedded youtube video

azangru:
This PR fixes link for an embedded youtube video (as [explained on StackOverflow](https://stackoverflow.com/a/25661346/3925302), links to youtube videos in iframes embedded in external sites should have pathname that contains the word `embed` instead of `watch`).

Before fix:

image

After fix:

image

Link to page in a sandbox:
http://ves-hx2-76.ebi.ac.uk:8410/info/genome/genebuild/haplotypes_patches.html


azangru:
@ens-st3 Unless I am missing something, these are regular links; they aren't iframe sources. Nothing extra required for them. As far as I can tell, there was only one embedded youtube video in public-plugins codebase.

PR #309, opened 2020-02-10
Move genome_annotation.html to index.html to include content in sidebar

leannehaggerty:
Following Anne's discovery that the Genebuild documentation was missing from the sidebar: 
moved the genome annotation notes to the file "index.html" (on Anne's advice)


ens-ap5:
Thanks, Leanne! I'll need to check for broken links before we push it all live, but this is a good start.

PR #311, opened 2020-02-25
Update urls for 1000 genomes ftp site

azangru:
Update ftp host name for 1000 ftp server.

Also, hack our way on the client where we talk with the search server at https://www.internationalgenome.org. The server has to receive ftp url as a parameter, returns empty response if we change the ftp url to the new one, and  is not going to be reconfigured to correctly handle the new one 😞 

**Jira:** ENSWEB-5524

**Sandbox:** http://ves-hx2-76.ebi.ac.uk:8410/info/docs/tools/index.html

~Related PR in ebi-plugins: https://github.com/Ensembl/ebi-plugins/pull/36~

### DIFF
--- a/tools/conf/SiteDefs.pm
+++ b/tools/conf/SiteDefs.pm
@@ -158,9 +158,10 @@ sub update_conf {
   $SiteDefs::GENOME_REST_FILE_URL  = "https://www.internationalgenome.org/api/beta/file/_search";
 
   #1000Genome tool variables
-  $SiteDefs::PHASE1_PANEL_URL   = "ftp://ftp.1000genomes.ebi.ac.uk/vol1/ftp/release/20110521/phase1_integrated_calls.20101123.ALL.panel";
-  $SiteDefs::PHASE3_PANEL_URL   = "ftp://ftp.1000genomes.ebi.ac.uk/vol1/ftp/release/20130502/integrated_call_samples_v3.20130502.ALL.panel";
-  $SiteDefs::PHASE3_MALE_URL    = "ftp://ftp.1000genomes.ebi.ac.uk/vol1/ftp/release/20130502/integrated_call_male_samples_v3.20130502.ALL.panel";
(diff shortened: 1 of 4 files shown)


ens-ma7:
Codes change are ok. Just a few clarification:
1) URL hasn't been updated in the documentation, i noticed this was asked in the jira ticket but couldn't find any reply, i guess the decision was to keep the old URL in the documentation?
2) The hack makes sense but clarification on this, will we revert back to the old URL after the DC migration because the comment says hack for DC migration? My understanding is that it will stay like this, not sure. If it is a permanent change, we need to make sure the old 1000G ftp URL is always kept working or else their API will fail.

ens-ma7:
Also i wanted to submit a job so that i can have a look at the parameters submitted in the database but i can't submit a job on the sandbox. @azangru few things to check if you have your sandbox setup properly:
1) check if you have the tools_hive plugin enable
2) If you do have the plugin enable, can you remove your MULTI.packed file and restart your sandbox.
3) Also can you please check beekeeper dev is running

azangru:
@ens-ma7 

I have made sure tools work in the sandbox (turned out ensembl-hive code got switched to master with the bulk git-ensembl --checkout master available command, and had to be switched back to version 2.2). Thanks to @jyothishnt for help.

ens-ma7:
I have just tried running a job for Data Slicer on your sandbox and it seems that the 1000Genome API is using the new FTP url and returning blank.

azangru:
@ens-ma7 it turned out that the ftp server ftp://ftp.ebi.ac.uk/1000g/ftp was misconfigured, and Ensembl server could not access files on it. Web prod has fixed the ftp server; so at least the sandbox works now. Don't know how high of a load that server can take.

ens-ma7:
Since this is a temporary fix for the DC migration, we need to make the PR against the release/99 branch and it will be reverted back in release/100.

ens-ma7:
codes change are ok and manage to run a job successfully with the right ftp parameters in the tools db. 
Once you make this PR against release/99 branch only, should be good.

azangru:
Changed the base branch to release/99 😃

PR #316, opened 2020-03-09
Fix banner overlap issue on the search page

msmanoj:
Description:
This PR is to fix the issue with the overlapping elements on the search page when an announcement banner is displayed.

Live URL:
http://grch37.ensembl.org/Multi/Search/Results?q=BRCA2;site=ensembl_all

Sandbox URL:
http://ves-hx2-76.ebi.ac.uk:42228/Multi/Search/Results?q=BRCA2;page=1

### DIFF
--- a/solr/htdocs/components/71_solr_search.css
+++ b/solr/htdocs/components/71_solr_search.css
@@ -174,8 +174,8 @@
 .se_query {
   background: none repeat scroll 0 0 white;
   padding: 16px 0;
-  position: fixed;
-  top: 80px;
+  position: absolute;
+  top: 0;
 }
 
 .quick_links {
--- a/solr/htdocs/solr/search_templates.css
+++ b/solr/htdocs/solr/search_templates.css
@@ -6,18 +6,18 @@
 .g_spinner {
   background: url("/i/e-loader.gif") repeat scroll 0 0 transparent;
   height: 66px;
-  left: 860px;
-  position: fixed;
-  top: 93px;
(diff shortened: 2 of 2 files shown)


tuanebi:
@msmanoj  The banner looks good now. However, the search does not work when I select one option from "Restrict category to" or "Restrict species to"

msmanoj:
@tuanebi That must be something to do with my sandbox setup and not this PR.

tuanebi:
@msmanoj Tested on a fallback machine and it worked. Thank you!

azangru:
The regressions I have noticed in the sandbox are:

- blank space above the sidebar that appears immediately upon scrolling:

image

- main content no longer scrolls independently of the sidebar.

Both regressions are probably minor. The first one is more noticeable. Despite the regressions, search page is more usable than it currently is in production, so we might as well push this change live. Not sure what the long-term strategy for the search page is — should we attempt to update its styles properly to both support the banner and preserve its former appearance/behaviour; or will we revert this change when we remove the banner?

One possible area to explore is to forcefully change all those `top` rules in appropriate elements when the banner is shown — either in the Perl code, if it's possible to get the information there, or in javascript.

msmanoj:
> blank space above the sidebar that appears immediately upon scrolling:

tuanebi:
@azangru Good points!
I think the 1st point is acceptable. The 2nd point might affect user experience, we can think of an improvement after DC migration

PR #323, opened 2020-04-08
Remove dN/dS from Compara docs (homology_method.html)

azangru:
Compara is removing dN/dS data from their database, and we are removing dN/dS from the site.

Related PR in ensembl-webcode: https://github.com/Ensembl/ensembl-webcode/pull/767

Jira ticket: ENSWEB-5457

Sandbox: http://ves-hx2-76.ebi.ac.uk:8410/info/genome/compara/homology_method.html

### DIFF
--- a/ensembl/htdocs/info/genome/compara/homology_method.html
+++ b/ensembl/htdocs/info/genome/compara/homology_method.html
@@ -57,8 +57,6 @@ given internal node of the tree.
 
 <li>A <a href="stables_id.html">stable ID</a> is assigned to each GeneTree.</li>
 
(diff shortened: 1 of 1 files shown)


azangru:
@ens-st3 do we also remove any mention about dN and dS here (same file):

image

ens-st3:
I suspect so but defer to @carlacummins

carlacummins:
The references in the tree building section should not be removed

azangru:
@carlacummins Do I understand it correctly that there should be no changes on [this page](http://www.ensembl.org/info/genome/compara/homology_method.html) at all?

carlacummins:
@azangru : Sorry - "Details on tree building" section, point 9. should be removed. "Tree building" section, points 4. and 5. should be kept. Thank you!

PR #324, opened 2020-04-16
Rename EPO_LOW_COVERAGE -> EPO_EXTENDED for e101

CristiGuijarro:
In the intentions for Compara e101 the `EPO_LOW_COVERAGE` method has been renamed to `EPO_EXTENDED` to better fit the functionality of the method. In this PR `EPO_LOW_COVERAGE` has been simply swapped to `EPO_EXTENDED` and `EPO-Low-Coverage` to `EPO-Extended`. With additional change in the docs to better describe `EPO_EXTENDED`.

Intention: [ENSINT-334](https://www.ebi.ac.uk/panda/jira/browse/ENSINT-334)


muffato:
Hi @CristiGuijarro 
There are also mentions with hyphens instead of underscores in `ensembl/htdocs/info/genome/compara/multiple_genome_alignments.html`

CristiGuijarro:
> Hi @CristiGuijarro
> There are also mentions with hyphens instead of underscores in `ensembl/htdocs/info/genome/compara/multiple_genome_alignments.html`

Thanks @muffato missed those! I have pushed those changes, but renamed using underscores instead of hyphens, is that okay?

muffato:
Looking at the page http://www.ensembl.org/info/genome/compara/multiple_genome_alignments.html it seems more natural to use the "display name", i.e. the same as would appear in the 3rd column of the table. What do you think ?

CristiGuijarro:
> Looking at the page http://www.ensembl.org/info/genome/compara/multiple_genome_alignments.html it seems more natural to use the "display name", i.e. the same as would appear in the 3rd column of the table. What do you think ?

I think display name is probably best, I agree, and have changed that now :+1: 
The previous 'convention' seems to be a little confused, so I wasn't sure :sweat_smile: 

Better now?

muffato:
Super, thank you !

PR #339, opened 2020-05-15
dots before doi removed

nds:


nds:
This combined two minor changes: some dots before the dois removed and a link to rfam added on the annotation page.

PR #342, opened 2020-05-19
Fix Windows new line chars

tuanebi:
There are RT tickets (417267, 416182) reporting an issue with PostGap. It turns out the users use Windows-style input file which has CRLF (\r\n) as new line chars. Our code only considers the Unix-style input (\n for new lines). 

This part here fails with Windows-style input because we expect `p-value` but Perl's chomp method give us `p-value\r`
https://github.com/Ensembl/public-plugins/blob/release/100/tools_hive/modules/EnsEMBL/Web/RunnableDB/Postgap.pm#L217-L221


ens-ma7:
@tuanebi have you checked if this works with linux style file where there is no \R or maybe you need your regex to not be case sensitive?

ens-ma7:
also @tuanebi the PR needs to be against postreleasefix/100 branch

tuanebi:
@ens-ma7 \R will match \r | \n | \r\n and also anything considered a linebreak sequence by Unicode, according to the doc here
https://perldoc.perl.org/perlrebackslash.html#Misc

I tested both Windows and Linux input format. 
My sandbox http://ves-hx2-75.ebi.ac.uk:9580/Homo_sapiens/Tools/Postgap so it would be best if you can test some inputs.

I'll create another PR to postreleasefix/100

tuanebi:
@ens-ma7 I changed the base branch of this PR from release/100 to postreleasefix/100

tuanebi:
mpush DONE: Successfully updated both branches with following commit(s):
e8e6985d9 Fix removing new line chars

PR #346, opened 2020-06-08
Fix sgd url

manuelcarbajo:
The bug was initially reported by Ameya and is described in the JIRA ticket:
https://www.ebi.ac.uk/panda/jira/browse/ENSINT-491
There has been a change in the syntax on how SGD is accessed
Instead of 
http://db.yeastgenome.org/cgi-bin/locus.pl?locus=###ID###
It needs to be
https://www.yeastgenome.org/locus/###ID###
 
There are at least two files where this information is stored:
https://github.com/Ensembl/public-plugins/blob/release/100/ensembl/conf/ini-files/saccharomyces_cerevisiae.ini#L70
and
https://github.com/EnsemblGenomes/eg-web-common/blob/release/eg/47/conf/ini-files/DEFAULTS.ini

Are there any others?

### DIFF
--- a/ensembl/conf/ini-files/saccharomyces_cerevisiae.ini
+++ b/ensembl/conf/ini-files/saccharomyces_cerevisiae.ini
@@ -69,10 +69,10 @@ ASSEMBLY_CONVERTER_FILES = [EF2_to_R64-1-1 EF3_to_R64-1-1 R64-1-1_to_EF2 R64-1-1
 
 [ENSEMBL_EXTERNAL_URLS]
 
-OPERON          = http://db.yeastgenome.org/cgi-bin/locus.pl?locus=###ID###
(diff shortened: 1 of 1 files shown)


ens-ap5:
Hi Manuel, it's difficult to find your changes because they're mingled in with loads of commits made by the webteam. Perhaps @msmanoj can comment on whether the PR needs cleaning up?

msmanoj:
Hi @ens-ap5, the branch was created from `master` but the base branch was mistakenly set to `release/100`. @manuelcarbajo confirmed that this fix is for 101 so I have changed the base branch to `master`.

msmanoj:
Hi @ens-ap5, could you please check if the changes Manuel has made needs to be done anywhere else?

ens-ap5:
> Hi @ens-ap5, could you please check if the changes Manuel has made needs to be done anywhere else?

I guess it might also need doing in eg-web-fungi

PR #354, opened 2020-06-26
COVID-19 - Fix FTP site URLs in the species homepage

msmanoj:
This PR is to fix the broken FTP site URLs in the COVID-19 site species homepage.

Related JIRA:
https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5872

Sandbox URL:
http://ves-hx2-76.ebi.ac.uk:42227/Sars_cov_2/Info/Index


msmanoj:
>That looks OK, thought the sprintf is rather superfluous if you're returning the string as-is (in the code it was copied from, extra characters are being inserted).

Ah yes! I'll remove it.

PR #357, opened 2020-07-01
Typo in href link

JAlvarezJarreta:
_Note:_ Not sure if there was a faster workaround, but this is the one I'm more familiar with.


muffato:
FYI, unless it's a bugfix for the current site, you don't need to make a pull request, just push directly

PR #376, opened 2020-09-03
Fix VEP column not found bug

msmanoj:
**Description:**
This PR is to fix the intermittent error that was breaking the VEP tool jobs.

The problem is something to do with the following line:
https://github.com/Ensembl/public-plugins/blob/release/100/tools/modules/EnsEMBL/Web/Job/VEP.pm#L155
Modifying the value of $param somehow gets temporarily changed in the stored config.
To fix this, I'm just cloning the $param defined in the foreach to $param_clone and using it instead.

**Example ticket that was throwing this error:**
http://debug.ensembl.org/Homo_sapiens/Tools/VEP/View?db=core;tl=9QezRSx98mS7heaW-6524731

**Related JIRA:**
https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5658

### DIFF
--- a/tools/modules/EnsEMBL/Web/Job/VEP.pm
+++ b/tools/modules/EnsEMBL/Web/Job/VEP.pm
@@ -144,20 +144,22 @@ sub _configure_plugins {
     
     foreach my $param(@{$plugin->{params}}) {
       
+      my $param_clone = $param;
+
       # links to something in the form
-      if($param =~ /^\@/) {
-        $param =~ s/^\@//;
+      if($param_clone =~ /^\@/) {
+        $param_clone =~ s/^\@//;
         
         my @matched = ();
         
         # fuzzy match?
-        if($param =~ /\*/) {
-          $param =~ s/\*/\.\*/;
-          @matched = grep {$_->{name} =~ /$param/} @{$plugin->{form}};
+        if($param_clone =~ /\*/) {
+          $param_clone =~ s/\*/\.\*/;
+          @matched = grep {$_->{name} =~ /$param_clone/} @{$plugin->{form}};
         }
         
         else {
-          @matched = grep {$_->{name} eq $param} @{$plugin->{form}};
+          @matched = grep {$_->{name} eq $param_clone} @{$plugin->{form}};
         }
         
         foreach my $el(@matched) {
@@ -176,7 +178,7 @@ sub _configure_plugins {
       
       # otherwise just plain text
       else {
-        push @params, $param;
+        push @params, $param_clone;
       }
     }
     


ens-ma7:
Hi @msmanoj sorry this change doesnt make sense, my $param should always be the same as param_clone so no need to reinitialise $param_clone. it might be something else the issue. we probably need to look into it together if you already have your sandbox setup for it.

msmanoj:
>my $param should always be the same

Hi @ens-ma7 , I thought the same but for some reason the actual value stored in the conf itself was getting changed temporarily. I tried changing the value of param to `123` and I when I logged the following value:
https://github.com/Ensembl/public-plugins/pull/376/files#diff-3e2121382b5955bcf6c5543e4d9c7231R131

for the next job, the value for param was `123`. I can reproduce this on the debug site. Cloning the $param seems to be fixing this.

ens-ma7:
Sorry @msmanoj  i still dont agree, you might be right but i am not convinced thats the correct solution, it might be something else. its probably best to have a chat about this.

ens-ma7:
Having gone through what the issue is, i understood where the problem is which is a nasty perl object referencing issue. I suggested a solution which we were not able to test. @msmanoj maybe you can try the temporary solution first and if it works push this instead but if it doesnt i am happy for you to push the temporary fix.

ens-ma7:
see if something like this works:

```
sub _configure_plugins {	sub _configure_plugins {
  my $self = shift;
  my $job_data = shift;
my $pl = [];
  	  
  # get plugin config into a hash keyed on key	  # get plugin config into a hash keyed on key
  $pl = $self->hub->species_defs->multi_val('ENSEMBL_VEP_PLUGIN_CONFIG');
```

msmanoj:
> see if something like this works:
> 
> ```
> sub _configure_plugins {	
>   my $self = shift;
>   my $job_data = shift;
> my $pl = [];
>   	  
>   # get plugin config into a hash keyed on key	  
>   $pl = $self->hub->species_defs->multi_val('ENSEMBL_VEP_PLUGIN_CONFIG');
> ```

Hi @ens-ma7 , I tried what you have suggested above and I was still able to reproduce the bug. I guess the current solution I got in this PR will be a good fix for now.

msmanoj:
Hi @ens-st3 , am I okay to get this merged?

ens-st3 (approved):

tuanebi:
I think this PR is safe to merge.
The `@{$plugin->{params}` gets updated when we change `$param`. And because of shared memory somehow, it randomly is re-used by subsequent requests.
Having a copy of `$param` will stop it

PR #381, opened 2020-09-21
102 core schema update

ameya1981:
Version updated to 102 for core schema documentation.


ens-bwalts (approved):
I re-generated diagrams for 102 and confirmed that they have not changed from 101. Sarcophilus harrisii is a new assembly without an assembly table for 102, so our one deletion is expected.

PR #385, opened 2020-10-05
Vep form update

ridwanamo:
Jira Ticket:
https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-5067

As part of the jira ticket to update the VEP form for Non-vertebrates; this is fixing the different options that needs to be visible/toggle on the vep form (for example some options such as CCDS should only be shown for human). 

sandbox URL: http://ves-hx-78.ebi.ac.uk:11002/Multi/Tools/VEP?db=core

### DIFF
--- a/tools/modules/EnsEMBL/Web/Component/Tools/VEP/InputForm.pm
+++ b/tools/modules/EnsEMBL/Web/Component/Tools/VEP/InputForm.pm
@@ -396,7 +396,7 @@ sub _build_identifiers {
   });
 
   $fieldset->add_field({
-    'field_class' => '_stt_Homo_sapiens _stt_Mus_musculus_stt_core _stt_merged _stt_gencode_basic',
+    'field_class' => '_stt_Homo_sapiens _stt_Mus_musculus _stt_core _stt_merged _stt_gencode_basic',
     'type'        => 'checkbox',
     'name'        => 'ccds',
     'label'       => $fd->{ccds}->{label},


tools/modules/EnsEMBL/Web/Component/Tools/VEP/InputForm.pm
```diff
@@ -396,13 +396,13 @@ sub _build_identifiers {
...
   $fieldset->add_field({
-    'field_class' => '_stt_core _stt_merged _stt_gencode_basic',
+    'field_class' => '_stt_Homo_sapiens _stt_Mus_musculus _stt_core _stt_merged _stt_gencode_basic',
     'type'        => 'checkbox',
     'name'        => 'ccds',
     'label'       => $fd->{ccds}->{label},
     'helptip'     => $fd->{ccds}->{helptip},
     'value'       => 'yes',
-  });
+  }) if (first { $_->{'value'} eq 'Homo_sapiens' } @$species);
```
jyothishnt, line 405: Missed Mus_musculus?

tools/modules/EnsEMBL/Web/Component/Tools/VEP/InputForm.pm
```diff
@@ -500,15 +500,17 @@ sub _build_variants_frequency_data {
...
-        'name' => 'failed',
-        'label' => $fd->{failed}->{label},
-        'helptip' => $fd->{failed}->{helptip},
-        'value' => 1,
-        'checked' => 0,
-        'field_class'   => [qw(_stt_yes _stt_allele)],
       })]
+    }) if (first { $_->{'value'} eq 'Homo_sapiens' } @$species);
+
+    $fieldset->add_field({
```
jyothishnt, line 506: This field is for Homo_sapiens only. Check screenshot.
Screenshot 2020-10-07 at 10 37 00

tools/modules/EnsEMBL/Web/Component/Tools/VEP/InputForm.pm
```diff
@@ -622,70 +624,72 @@ sub _build_additional_annotations {
...
-    my @cell_types = ();
-    foreach (sort {$a->short_name cmp $b->short_name} @{$regulatory_build->get_all_Epigenomes}) {
-      my $short_name = $_->short_name;
-      my $rm_white_space_label = $short_name;
-      $rm_white_space_label =~ s/ /\_/g;
-      push @cell_types, { value => $rm_white_space_label, caption => $short_name }; 
-    } 
+    $fieldset = $form->add_fieldset({'legend' => $current_section, 'no_required_notes' => 1, class => $regu_class});
 
-    $fieldset->add_field({
```
jyothishnt, line 647: Hard to understand this change. Probably a small thing but could we do it over a call if it is hard to explain it here?

tools/modules/EnsEMBL/Web/Component/Tools/VEP/InputForm.pm
ens-ma7: Yep thats correct, its been there before and it is specific to human only.

tools/modules/EnsEMBL/Web/Component/Tools/VEP/InputForm.pm
ens-ma7: Fall in the same category as vertebrates so we only need to check for human or mouse, both is not needed

jyothishnt: wonder why did it have _stt_mus_musulus mentioned in the class before?

ens-ma7: in the class it is the correct way because it has to toggle between mouse and human but the if check in the perl is to add them to the form first. Remember the fix we discuss to generate the _stt toggle data for each species.

jyothishnt (approved):
Looks good.

PR #388, opened 2020-10-13
Remove hiding of alternate vep output fields

aparton:
There are some links from other parts of the documentation that link to hidden sections on vep_formats.html (e.g. info/docs/tools/vep/script/vep_other.html#refseq_match). 

Unhiding the section removes this issue, and makes it easier for users to find the documentation required.


dglemos:
Tested with my sandbox. It looks good.

dglemos (approved):

PR #393, opened 2020-10-30
ENSWEB-5921: Fix wrong example FTP URL

imransl:
The GRCh37 site's help page 'Installing the Ensembl Data' shows the vertebrate (www) FTP URL. This PR fixes that by adding a SiteDefs variable.

This PR needs to be merged at the same time as [68 in ebi-plugins](https://github.com/Ensembl/ebi-plugins/pull/68) is merged.

### DIFF
--- a/ensembl/conf/SiteDefs.pm
+++ b/ensembl/conf/SiteDefs.pm
@@ -32,7 +32,6 @@ sub update_conf {
   $SiteDefs::ARCHIVE_BASE_DOMAIN        = 'archive.ensembl.org';
   $SiteDefs::ENSEMBL_REST_URL           = 'https://rest.ensembl.org';  # URL for the REST API
   $SiteDefs::ENSEMBL_REST_DOC_URL       = 'https://github.com/Ensembl/ensembl-rest/wiki';
-  $SiteDefs::ENSEMBL_EXAMPLE_FTP_URL    = 'ftp://ftp.ensembl.org/pub/current_mysql';
 
   $SiteDefs::GXA                        = 1; #enabling gene expression atlas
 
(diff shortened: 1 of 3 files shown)


ens-ap5 (approved):

ensembl/conf/SiteDefs.pm
```diff
@@ -32,6 +32,7 @@ sub update_conf {
   $SiteDefs::ARCHIVE_BASE_DOMAIN        = 'archive.ensembl.org';
   $SiteDefs::ENSEMBL_REST_URL           = 'https://rest.ensembl.org';  # URL for the REST API
   $SiteDefs::ENSEMBL_REST_DOC_URL       = 'https://github.com/Ensembl/ensembl-rest/wiki';
+  $SiteDefs::ENSEMBL_EXAMPLE_FTP_URL    = 'ftp://ftp.ensembl.org/pub/current_mysql';
```
ens-ma7, line 35: If I understand the issue, you need a parameter for the mysql FTP URL. we already have a parameter for FTP URL (ENSEMBL_FTP_URL). so maybe the new parameter should be:
ENSEMBL_MYSQL_FTP_URL = ENSEMBL_FTP_URL."/current_mysql"; 
and change it accordingly in the 37 plugins

From what i understand you should probably have two parameters:

imransl:
I have renamed the URL variable and also added it into DEFAULTS.ini instead of SiteDefs.pm

ens-ma7:
see my comments in the other PR,  in here the value is correct but i still feel the naming should be ENSEMBL_MYSQL_FTP_URL since it is specific to the mysql ftp url

imransl:
@ens-ma7 Sorry I got carried looking at the line above it. There is already a variable for its HTTP equivalent and it looks like this: `ENSEMBL_FTP_OVER_HTTP_URL = http://ftp.ensembl.org/pub`

I'll rename it to `ENSEMBL_MYSQL_FTP_URL`.

imransl:
@ens-ma7 I've renamed it in both the repositories

ens-ma7 (approved):

ens-ma7:
@imransl thanks for changing the parameter name, the parameter above make sense the naming. The changes are ok and i am assuming it is working fine on your sandbox(please do check this before merging, i couldn't check it because there was no sandbox URL).

ens-ap5 (approved):

PR #412, opened 2021-01-06
Fix variant synonyms in web vep

dglemos:
Fix https://github.com/Ensembl/public-plugins/pull/404 
Sandbox: http://ves-hx2-76.ebi.ac.uk:7040/Homo_sapiens/Tools/VEP?tl=hix2ELsDmBobb5II-392


ens-ma7:
Definitely not expecting this change to fix it, I thought it needed something in the javascript file. @dglemos how did you figure out by changing it to yes it will work.
I have tested the change and it works, happy for it to merge but just curious how that fix works.

ens-ma7 (approved):

dglemos:
@ens-ma7 I compared other options and the ones with 'yes' work properly. I just tried to change the value and it worked. I don't know if there's any javascript, probably there is something.

aparton (approved):

PR #416, opened 2021-01-15
Variant Recoder - update documentation 

dglemos:
Variant recoder is also available as a web tool.
Sandbox: http://ves-hx2-76.ebi.ac.uk:7040/info/genome/variation/tools/variant_tools.html

### DIFF
--- a/ensembl/htdocs/info/genome/variation/tools/variant_tools.html
+++ b/ensembl/htdocs/info/genome/variation/tools/variant_tools.html
@@ -89,6 +89,7 @@
           </div>
           <div style="float:right">
             <span style="display:table-cell">Availability:</span>
+            <a href="/Tools/VR" class="_ht" style="display:table-cell" title="Web tool"><img src="/img/vep_globe.png" style="border:1px solid #CCC;border-radius:5px;background-color:#336;width:48px;vertical-align:middle;margin-left:15px"/></a>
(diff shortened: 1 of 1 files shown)


ens-st3:
Will the tool work for GRCh37 ?

ens-ma7:
Hi @dglemos its a good point that @ens-st3 made because right now by default VR will be shown on all sites/divisions since it is enabled in the main tools/conf/SiteDefs.pm

If it is available for grch37 then we will need to disable it for grch37 like this:
https://github.com/Ensembl/ebi-plugins/blob/release/102/grch37/conf/SiteDefs.pm#L90 (replace _PG with _VR)

dglemos:
It also should be available on grch37.

ens-ma7:
thanks @dglemos then we dont have to do anything in the code.

ens-ma7 (approved):

ens-ma7:
@ens-st3 just waiting for you to confirm if its ok with you the changes or is there anything else, then we can merge the PR.

ens-st3:
This can be merged then :-)

Is there a specific config file for GRCh37 (since it'll use a different DB) ?

ens-ma7:
The way the settings has been configured it should pickup the default database for each site/division. But will be good to have a look at the settings inside the tools database once test is up. @dglemos please make note we should test both test.ensembl.org and test-grch37.ensembl.org and look at the settings in the tools database.

PR #421, opened 2021-02-01
Fix VEP file url input

msmanoj:
Related JIRA:
https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6151

Description:
- This PR fixes the issue with VEP file url validation issue mentioned in the related ticket 

TODO: Need to cherry-pick the changes onto `postreleasefix/103`

Sandbox URL:
http://ves-hx2-75.ebi.ac.uk:42229/Tools/VEP

### DIFF
--- a/tools/htdocs/components/20_VEPForm.js
+++ b/tools/htdocs/components/20_VEPForm.js
@@ -107,7 +107,7 @@ Ensembl.Panel.VEPForm = Ensembl.Panel.ToolsForm.extend({
         $(this).data('valid', false);
         return;
       }
-      if ($.trim($(this).find('.vep-input').val()) == "" && $(this).find('.ffile').val() == "" && $(this).find('.ftext._url ').val() == "" ) {
+      if ($.trim($(this).find('.vep-input').val()) == "" && $(this).find('.ffile').val() == "" && $(this).find('.ftext.url ').val() == "" ) {
         panel.showError('Please provide an input.', 'No input found');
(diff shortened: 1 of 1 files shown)


tools/htdocs/components/20_VEPForm.js
```diff
@@ -107,7 +107,7 @@ Ensembl.Panel.VEPForm = Ensembl.Panel.ToolsForm.extend({
         $(this).data('valid', false);
         return;
       }
-      if ($.trim($(this).find('.vep-input').val()) == "" && $(this).find('.ffile').val() == "" ) {
+      if ($.trim($(this).find('.vep-input').val()) == "" && $(this).find('.ffile').val() == "" && $(this).find('.ftext._url ').val() == "" ) {
```
jyothishnt, line 110: Hi @msmanoj  This should work. 

But I have heard that using underscores are like "_url"  denotes a private variable. Here is the class assigned when we create the dom element here https://github.com/Ensembl/public-plugins/blob/release/102/tools/modules/EnsEMBL/Web/Component/Tools/VEP/InputForm.pm#L198

How about using "url" instead of "_url" just for safer side?

msmanoj: @jyothishnt - Yes, that makes sense. I actually didn't notice that it. has `url` as well. I've updated it

jyothishnt (approved):

PR #423, opened 2021-02-02
VEP doesn't work for mouse strains

tuanebi:
Fixing an issue that VEP doesn't work for mouse strains [ENSWEB-6150](https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6150)

Web form submits `species_url` but VEP expects `production_name` when searching VEP cache. Ignoring letter case, `species_url` and `production_name` are the same for most species, so VEP is still working for most species. For mouse strains (and possibly some others?), the two values are different, and VEP fails.

Working on sandbox:
http://ves-hx2-75.ebi.ac.uk:9580/Multi/Tools/VEP?db=core


ens-st3 (approved):

tuanebi:
Made a manual merge

PR #424, opened 2021-02-02
update vep examples mouse strains prf103

at7:
update VEP examples for mouse strains


aparton (commented):
All of the rsIDs have been removed, is this intentional? VEP_PILEUPs are all missing as well

tuanebi:
I tested these strains with rsIDs input, both VEP and instant VEP failed because of no variation database. So I did the same to remove VEP_ID from my PR https://github.com/Ensembl/public-plugins/pull/423

at7:
When the ini files for mouse strains were first created the VEP input data was just copied from mouse. I now ran the generate_vep_examples script for the mouse strains. Mouse strains don't have variation databases and hence no rsIDs. And we don't support pileup format anymore, at least it isn't an option on the tools interface.

aparton (approved):
Fair enough, then I'm happy with this. Tested a few random variants and they're valid.

tuanebi:
Thank you @at7 and @aparton

PR #440, opened 2021-03-03
Update DisGeNET output (104)

dglemos:
DisGeNET has a different output structure, `DisGeNET_PMID`, `DisGeNET_SCORE` and `DisGeNET_disease` are no longer used. Instead, the header is just `DisGeNET`. 
Sandbox to test: http://ves-hx2-76.ebi.ac.uk:7040/Tools/VEP 
input variant: rs699

### DIFF
--- a/tools/modules/EnsEMBL/Web/Component/Tools/VEP/Results.pm
+++ b/tools/modules/EnsEMBL/Web/Component/Tools/VEP/Results.pm
@@ -145,6 +145,22 @@ sub content {
   my $headers = $header_hash->{'combined'};
   my $header_extra_descriptions = $header_hash->{'descriptions'} || {};
 
+  # Overwrite DisGeNET header description
+  if(exists $header_extra_descriptions->{'DisGeNET'}) {
+    my %new_header_extra_descriptions;
+    foreach my $key (keys %$header_extra_descriptions) {
+      if($key eq 'DisGeNET') {
+        my $description = $header_extra_descriptions->{$key};
+        $description =~ s/Each value is separated.*//;
+        $new_header_extra_descriptions{$key} = $description;
+      }
+      else {
+        $new_header_extra_descriptions{$key} = $header_extra_descriptions->{$key};
+      }
+    }
+    $header_extra_descriptions = \%new_header_extra_descriptions;
+  }
+
   my $actual_to = $from - 1 + ($line_count || 0);
   my $row_count = scalar @$rows;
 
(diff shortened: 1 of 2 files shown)


tools/modules/EnsEMBL/Web/Component/Tools/VEP/Results.pm
```diff
@@ -145,6 +145,22 @@ sub content {
   my $headers = $header_hash->{'combined'};
   my $header_extra_descriptions = $header_hash->{'descriptions'} || {};
 
+  # Overwrite DisGeNET header description
+  if(exists $header_extra_descriptions->{'DisGeNET'}) {
```
at7, line 149: Do you have an example of what the description looks like?
You can also update the value like this:
$header_extra_descriptions->{'DisGeNET'} =~ s/Each value is separated.*//;

at7 (commented):
Overall the PR looks good. I'm only interested in if we can shorten updating the description.

tools/modules/EnsEMBL/Web/Component/Tools/VEP/Results.pm
dglemos: I added an example of the description. The last part should be deleted because the output has a different structure and the values are not separated by ':'.

tools/modules/EnsEMBL/Web/Component/Tools/VEP/Results.pm
```diff
@@ -145,6 +145,15 @@ sub content {
   my $headers = $header_hash->{'combined'};
   my $header_extra_descriptions = $header_hash->{'descriptions'} || {};
 
+  # Overwrite DisGeNET header description
+  # Description example: "Variant-Disease-PMID associations from the DisGeNET database. The output includes 
+  #   the PMID of the publication reporting the Variant-Disease association, DisGeNET score for the Variant-Disease association, 
+  #   name of associated disease. Each value is separated by ':'"
+  # In Web VEP, the values are not separated by ':'
+  if(exists $header_extra_descriptions->{'DisGeNET'}) {
+    $header_extra_descriptions->{'DisGeNET'} =~ s/Each value is separated.*//;
```
at7, line 154: Thank you for the description example. There is an extra space which you also need to match in your regex before **Each ...**.  And just to clarify the the comment I would add: In Web VEP, the values are not separated by ':' **and therefore remove it from the description.**

at7 (approved):
Looks good in the sandbox and the code changes look good too

PR #442, opened 2021-03-16
Update Core Schema Diagram for Release 104

TamaraNaboulsi:
Core schema diagram update for release 104


ens-st3 (approved):
HTML is correct. Will assume the data is as this is a routing process.

PR #459, opened 2021-07-21
Variant Recoder - add options to report variant synonyms and MANE transcripts

dglemos:
1. Add two options to Variant Recoder:
    - report variant synonyms 
    - report MANE select transcripts (in HGVS format)

2. Fix bug:  if the option `VCF string` isn't selected in the input form, then vcf output file is empty.
3. Set up Variant Recoder to also use VEP examples from ini files

release 105
Sandbox: http://ves-hx2-76.ebi.ac.uk:7040/Homo_sapiens/Tools/VR


tools/modules/EnsEMBL/Web/Component/Tools/VR/Results.pm
```diff
@@ -319,6 +360,110 @@ sub linkify {
...
+
+  my @all_values = split(', ', $values);
+
+  foreach my $value (@all_values) {
+    return '-' unless defined $value && $value ne '';
+
+    my $mane_result;
+    my ($hgvsg, $hgvsc, $hgvsp) = split /\;/, $value;
+
+    if($hgvsc =~ /^ENST/) {
```
aparton, line 393: I'm 99% sure the answer to this question is no, but do we have any mane transcripts that don't begin with ENST?

tools/modules/EnsEMBL/Web/Component/Tools/VR/Results.pm
```diff
@@ -319,6 +360,110 @@ sub linkify {
...
+      my $zmenu_url = $hub->url({
+        type    => 'ZMenu',
+        action  => 'Transcript',
+        t       => $split_value[0],
+        species => $species,
+        db      => $db_type,
+      });
+
+      my $transp_value = $self->zmenu_link($url, $zmenu_url, $split_value[0]);
+      $mane_result .= "&nbsp;(".$transp_value.":".$split_value[1].")";
```
aparton, line 438: I don't like the design of adding the protein level description as an addition in brackets, it looks a little forced. Probably worth (another, sorry!) a team-wide discussion about exactly how we want this to look.

tools/modules/EnsEMBL/Web/Object/VR.pm
```diff
@@ -107,7 +107,7 @@ sub get_form_details {
     $self->{_form_details} = {
         id => {
           'label'   => 'Variant identifier',
-          'helptip' => 'dbSNP rsIDs or any synonym for a variant present in the Ensembl Variation database',
+          'helptip' => 'Variants present in the Ensembl Variation database that are co-located with input',
```
aparton, line 110: Can we simplify this tooltip? I suggest 'Known co-located variants'. I don't think we get any value from referencing the ensembl variation database

tools/modules/EnsEMBL/Web/Object/VR.pm
```diff
@@ -129,6 +129,14 @@ sub get_form_details {
           'label'   => 'VCF format',
           'helptip' => 'Position based name',
         },
+        var_synonyms => {
+          'label'   => 'Variant synonyms',
+          'helptip' => 'Extra known synonyms for co-located variants',
```
aparton, line 134: Suggest changing 'extra' to 'additional'

aparton (commented):
Looks good. Core functionality seems to work, all of the synonyms links come out correctly formatted, MANE select looks good.  Couple of minor comments:

tools/modules/EnsEMBL/Web/Component/Tools/VR/Results.pm
ens-ap5: No, because MANE only applies to human and all human transcripts will have an ENST stable id. From the NCBI website: "The goal of this project is to provide a minimal set of matching RefSeq and Ensembl transcripts of human protein-coding genes"

tools/modules/EnsEMBL/Web/Component/Tools/VR/Results.pm
dglemos: Adding brackets was the only way I found to show the HGVS protein without being too much but it's not the best display. We can have a chat about the display on the next team meeting.

tools/modules/EnsEMBL/Web/Object/VR.pm
dglemos: I tried to follow this description we have https://www.ensembl.org/info/docs/tools/vep/vep_formats.html#id 
What about `Identifier of any existing co-located variants`?

sarahhunt:
I know this is an usual use, but if I enter 2 identifiers for the same variant, only the first one gets MANE info. Example input: 

- VCV000018068 & PA166153539
-  PA166153539 & rs699

dglemos:
@sarahhunt I opened a ticket to investigate further.

sarahhunt:
Thanks Diana. That's an edge case people are unlikely to hit, so this looks good.

sarahhunt (approved):

PR #462, opened 2021-08-10
Feature/viral collections

ens-ap5:
Support for covid19 collection databases - needs testing on a staging site.


ens-ap5:
These changes only affect the covid plugin, and will not affect the current covid site since that's on release/101. Therefore no review is required.

PR #469, opened 2021-09-12
Fix syntax in Panel.js file in genoverse public plugin

azangru:
I've been playing with a new google closure compiler, when it complained of an error in the merged javascript:

```
/homes/ens_adm06/tmp/8410/server/minified/11f009db9960d736e3a9413cab8870af.js.tmp:9396:34: ERROR - [JSC_PARSE_ERROR] Parse error. invalid assignment target                                               
  9396|               track.prop('order') = orderReverse; 
```

The error must be coming from [here](https://github.com/Ensembl/public-plugins/blob/release/104/genoverse/htdocs/genoverse/ensembl/js/Panel.js#L505), and the compiler is correct — it is not valid javascript. This looks like a bit of jQuery, where the correct syntax will be as in this PR.

I don't know anything about the genoverse plugin, so couldn't check where this code is expected to run. Does the change make sense to you?

### DIFF
--- a/genoverse/htdocs/genoverse/ensembl/js/Panel.js
+++ b/genoverse/htdocs/genoverse/ensembl/js/Panel.js
@@ -502,7 +502,7 @@ Ensembl.Panel.Genoverse = Ensembl.Panel.ImageMap.extend({
           
           if (track && track.prop('unsortable') !== true) {
             if (track.prop('strand') === -1 && orderReverse) {
-              track.prop('order') = orderReverse;
+              track.prop('order', orderReverse);
               track.prop('label').data('order', orderReverse);
               
               track = track.prop('forwardTrack');


ens-ap5:
I don't think this code is being run, for two reasons:

1. If you grep for "unsortable" in public-plugins/genoverse, you'll see that it's set to true in a great many places.
2. The code was written by Simon to be somewhat standalone and we don't use all of its features - for example we don't separate the features by strand, so I'm not sure this reverse sorting on the reverse strand is relevant

So, feel free to fix the syntax if it prevents pointless warnings, but I don't think there's any way to test it in the browser.

ens-ma7 (approved):

ens-ma7:
As @ens-ap5 said we are probably not using it anywhere thats why nothing is broken :) And your change you made looks the right thing to do.

azangru:
Ok, I'll go ahead and merge it. Master branch should be ok, right?

ens-ma7:
Yes master branch should be fine.

PR #476, opened 2021-09-20
Add 3D viewer for Alphafold-predicted protein structures

azangru:
# Description
This PR adds a viewer of 3D protein structures predicted by the AlphaFold AI to Ensembl pages. The first organism that we are using to test the Alphafold integration is wheat (Arabidopsis thaliana).

The proposed widget incorporates the [pdbe-molstar plugin](https://github.com/PDBeurope/pdbe-molstar) and wraps it into a custom web component using the [Lit](https://lit.dev) library for convenience. The code is written in modern javascript (supported by ~95% of web browsers) and saved into `/widgets/htdocs/alphafold`, outside of the `components` folder which would be processed by our regular build scripts.

**Note** that because this code exists our usual build pipeline, the files will not be concatenated, minified and stamped with a hash in a usual way. To avoid caching issues, this PR disables caching of the static files in the `alphafold` folder. Since they are tiny, this should not result in significant performance issues.

## Sandbox
http://ves-hx2-75.ebi.ac.uk:8450/Arabidopsis_thaliana/Transcript/AFDB?db=core;g=AT3G52430;r=3:19431095-19434450;t=AT3G52430.1

The pull request is submitted against the `master` branch.


ens-st3 (commented):
Functionally the widget behaves as I'd expect.
Geneview pages need fixing on this sandbox though before I can approve

azangru:
> Geneview pages need fixing on this sandbox though before I can approve

I believe they are fixed now, @ens-st3

jgtate (approved):
Looks good to me.

widgets/htdocs/alphafold/README.md
```diff
@@ -0,0 +1,14 @@
+# Description
+This folder contains code for a widget to display 3d protein structures predicted by the Alphafold AI.
+
+The widget uses `pdbe-molstar` as a viewer of 3D protein molecules, and wraps it in a UI relevant for Ensembl pages. The widget is built as a custom web component, using the `lit` library for convenience (to enable reactive properties on the HTML element). The code is written in modern javascript syntax that should work the browsers that support `<script type="module" />` (around 95% of web browsers).
+
+The widget is loaded from the `htdocs/components/95_AFDB.js` file. The top-level component (see `index.js`) is does not have shadow DOM, because shadow DOM prevents events from leaving the component, whereas the `pdbe-molstar` viewer uses an older version of React, whose events need to bubble up all the way to the `document` object to be registered. As a consequence of that, the CSS required by the top-level component is placed in `95_AFDB.css`, along with the relevant CSS variables for theming. Other components use shadow DOM, and contain their own styles.
```
ens-ma7, line 6: `is does` -> `does` (there is an extra is) in the first line

ens-ma7: i didnt quite understand the part of shadow DOM, something we can discuss during the review

widgets/htdocs/alphafold/controlPanel.js
```diff
@@ -0,0 +1,38 @@
+import { html, css, LitElement } from 'https://unpkg.com/lit@2.0.0-rc.4/index.js?module';
```
ens-ma7, line 1: i dont understand whats happening here?

widgets/modules/EnsEMBL/Web/Component/Transcript/AFDB.pm
```diff
@@ -0,0 +1,63 @@
...
+use strict;
+
+use HTML::Entities qw(encode_entities);
+use URI::Escape;
+
+use base qw(EnsEMBL::Web::Component::Transcript EnsEMBL::Web::Component::AFDB);
+
+sub _init {
+  my $self = shift;
+  $self->cacheable(0);
```
jyothishnt, line 31: Should we cache it? If the page is heavy/slow to load then we should probably think of caching?

widgets/modules/EnsEMBL/Web/Apache/Static.pm
```diff
@@ -0,0 +1,36 @@
...
+    # default behaviour
+    my $thirty_days = 60 * 60 * 24 * 30;
+    $r->headers_out->set('Cache-Control'  => 'max-age=' . $thirty_days);
+    $r->headers_out->set('Expires'        => HTTP::Date::time2str(time + $thirty_days));
+  }
+}
+
+sub should_skip_caching {
+  my $uri = shift;
+  my @patterns = (
```
ens-ma7, line 22: why an array? If it is just one string regex, no need for an array and this will get rid of the foreach below.

ens-ma7: i will write this like this:
{code}
my $uri = shift;

if($uri =~ qr/alphafold\/.*\.js$/) {
 return 1;
} else {
 return 0;
}
 {code}

widgets/htdocs/alphafold/controlPanel.js
jyothishnt: thinking of this, what if this module is not available from that cdn or if the cdn is down?
Do you think having this library locally would be good?

widgets/modules/EnsEMBL/Web/Component/AFDB.pm
```diff
@@ -0,0 +1,78 @@
+=head1 LICENSE
```
ens-ma7, line 1: I dont understand why you need this module (you can put all of this in the module `widgets/modules/EnsEMBL/Web/Component/Transcript/AFDB.pm` ). Maybe something to discuss in the group review

widgets/modules/EnsEMBL/Web/Component/AFDB.pm
```diff
@@ -0,0 +1,78 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+
+=cut
+
+package EnsEMBL::Web::Component::AFDB;
+
+use strict;
+
+use base qw(EnsEMBL::Web::Component::Shared);
```
ens-ma7, line 24: not sure you need this!

widgets/modules/EnsEMBL/Web/Component/AFDB.pm
ens-ma7: so the only reason you would have this module is if you want to have the alphafold widget on a different page/component but it seems its used only one place.

widgets/modules/EnsEMBL/Web/Component/Transcript/AFDB.pm
```diff
@@ -0,0 +1,63 @@
...
+  $self->ajaxable(1);  
+}
+
+sub content {
+  my $self        = shift;
+
+  my $hub         = $self->hub;
+  my $object      = $self->object;
+  my $html;
+  if ($object->Obj->isa('Bio::EnsEMBL::Transcript')) {
```
ens-ma7, line 41: i dont think you need this check. You are already in a transcript component.

widgets/htdocs/alphafold/controlPanel.js
azangru: > i dont understand whats happening here?

A third-party library is imported.

> thinking of this, what if this module is not available from that cdn or if the cdn is down? Do you think having this library locally would be good?

Unkpg is powered by Cloudflare (see [their home page](https://unpkg.com)); should be fairly robust.

widgets/modules/EnsEMBL/Web/Apache/Static.pm
azangru: > why an array?

Arrays are easy to extend. If we refactor the pdbe widget in a similar manner (there are plans to do so); we would also add the pdbe folder to the patterns list. Are there any problems with using the array?

widgets/modules/EnsEMBL/Web/Component/AFDB.pm
azangru: I just mirrored what has been done for the pdbe widget on the perl side.

widgets/modules/EnsEMBL/Web/Component/Transcript/AFDB.pm
azangru: Ok. I believe this is mostly just copy-pasting from the PDBE module.

widgets/modules/EnsEMBL/Web/Component/AFDB.pm
azangru: I deleted it, and nothing blew up; so you are right 👍

widgets/modules/EnsEMBL/Web/Component/Transcript/AFDB.pm
azangru: Deleted the check.

widgets/htdocs/alphafold/controlPanel.js
ens-ma7: sorry i meant the whole file, what does it do?

widgets/modules/EnsEMBL/Web/Apache/Static.pm
ens-ma7: array makes the code slower. if takes less memory not that it makes a big difference here but also the code is cleaner without the array (easier to understand). if we decide to add pdbe we can just add it to the regex.

widgets/htdocs/alphafold/controlPanel.js
azangru: It's encodes the visual appearance of a panel as a common component. A block with a title and a body. The styles for the title are the same across all such panels (exons, protein features, variants). What's different is what goes into the body; and that is defined in each specific panel file (exonsControlPanel, variantsControlPanel, etc.)

widgets/htdocs/alphafold/colorHelpers.js
```diff
@@ -0,0 +1,35 @@
+const commonColorsToRGBMap = {
```
ens-ma7, line 1: why do we need all these colour mappings? i get the RGB one but why the Hex one as well. There is a mixture of how the colours are used (its either HEX or RGB)

widgets/htdocs/alphafold/controlPanel.js
```diff
@@ -0,0 +1,38 @@
...
+    return html`
+      <div class="panel-title">
+        ${this.title}
+      </div>
+      <slot></slot>
+    `;
+  }
+};
+
+customElements.define('control-panel', ControlPanel);
```
ens-ma7, line 38: where is this used?

widgets/modules/EnsEMBL/Web/Apache/Static.pm
ens-ma7: i dont think you need a function to should_skip_caching. just have the code like this:
`if($uri =~ /alphafold\/.*\.js$/) {` instead of the function. By having this in the plugin, you are already expecting to put whatever is needed to skip the caching in the function. Creating another function here is just adding complexity to the code.

widgets/modules/EnsEMBL/Web/Apache/Static.pm
```diff
@@ -0,0 +1,36 @@
+package EnsEMBL::Web::Apache::Static;
+
+use strict;
+use warnings;
+
+sub add_caching_headers {
```
jyothishnt, line 6: decided to handle this using $self->PREV::add_caching_headers in code review.

widgets/modules/EnsEMBL/Web/Apache/Static.pm
jyothishnt: In code review, we decided to use this method bcs it is more readable... with some minor additions like below to create the match strings.
(
'aplhafold',
pdbe'
) => () => { to return a string for matching }

ens-ma7:
From code review meeting also to do:
- handle rest call failure (Show error message)
- Get rid of extra component (`Component/AFDB.pm`)

widgets/htdocs/alphafold/controlPanel.js
azangru: This associates the `control-panel` html tag with the `ControlPanel` class in web browser's registry of custom elements.

widgets/modules/EnsEMBL/Web/Apache/Static.pm
azangru: Discovered that we can't. I extracted a method from `add_caching_headers` specifically to check whether the response to a particular url needs to be cached, and am overriding it in public plugins. The logic that adds caching headers will remain unmodified in `ensembl-webcode`.

azangru:
> - handle rest call failure (Show error message)
> - Get rid of extra component (Component/AFDB.pm)

Done for both.

widgets/modules/EnsEMBL/Web/Apache/Static.pm
azangru: With some modifications, implemented here: https://github.com/Ensembl/public-plugins/blob/7f9fbb0a27df59b6c6aae725d11449c12f13d89c/widgets/modules/EnsEMBL/Web/Apache/Static.pm

widgets/modules/EnsEMBL/Web/Apache/Static.pm
```diff
@@ -0,0 +1,19 @@
...
+
+my @patterns_to_skip_caching = (
+  qr/alphafold\/.*\.js$/ # javascript files in the alphafold folder
+);
+
+sub should_skip_caching {
+  my $r = shift;
+  my $uri = $r->uri;
+
+  return any { $uri =~ $_ } @patterns_to_skip_caching;
```
ens-ma7, line 15: i dont think you need `any` , it will return 1 if the pattern match

widgets/htdocs/components/95_AFDB.js
```diff
@@ -0,0 +1,101 @@
...
+
+  onScriptLoaded: function() {
+    var enspIdElement = document.querySelector('#ensp_id'); // <-- expecting 1 or 0 HTML input elements
+    if (!enspIdElement) {
+      this.onMissingData();
+      return;
+    }
+
+    var container = document.querySelector('#alphafold_container');
+    var ensemblAlphafoldElement = document.createElement('ensembl-alphafold-viewer');
```
ens-ma7, line 50: I am just thinking instead of creating the element here, you can create it already in the perl file and instead of using the extra setAttribute to set the data attribute, you can set it already in the perl side. This way it keeps the code cleaner and also avoid an extra layer of converting the data (perl -> jquery -> web-component). And in the future if you need an extra parameter, you can just add it from the perl module. Does it make sense?

widgets/modules/EnsEMBL/Web/Component/Transcript/AFDB.pm
```diff
@@ -0,0 +1,103 @@
...
+}
+
+sub get_rest_urls {
+  my $self = shift;
+  my $hub  = $self->hub;
+
+  my $ensembl_rest_url = $hub->species_defs->ENSEMBL_REST_URL;
+
+  return qq{
+    <input class="panel_type" value="AFDB" type="hidden" />
```
ens-ma7, line 70: I would move this line to the main content function as it is not related to the rest_urls and it is also a very important part of the main content to initialise the panel.

widgets/modules/EnsEMBL/Web/Apache/Static.pm
azangru: I'm confused; my `@patterns_to_skip_caching` are still an array...

widgets/htdocs/components/95_AFDB.js
azangru: Yes, fair point. I have moved the code for setting up the web component to the server.

widgets/modules/EnsEMBL/Web/Component/Transcript/AFDB.pm
azangru: Combined everything into a single function. It turned out conveniently small.

widgets/htdocs/alphafold/exonsControlPanel.js
```diff
@@ -0,0 +1,170 @@
...
+        <div class="panel-summary">
+          Exons
+          <span class="exons-count">${this.exons.length}</span>
+          <button
+            class="show-feature ${this.areAllExonsVisible() ? 'show-feature_visible' : 'show-feature_hidden'}"
+            @click=${this.toggleAllExons.bind(this)}
+          ></button>
+          <button
+            class="panel-content-toggle ${this.isExpanded ? 'toggle-expanded' : 'toggle-collapsed'}"
+            @click=${() => this.isExpanded = !this.isExpanded}
```
jyothishnt, line 125: doubt, can you please tell me why @click is used instead of onClick?

widgets/htdocs/alphafold/index.js
```diff
@@ -0,0 +1,245 @@
...
+      proteinFeatures
+    };
+  }
+
+  initializeMolstar(afdbId) {
+    this.molstarInstance = new PDBeMolstarPlugin();
+
+    const options = {
+      customData: {
+        url: `https://alphafold.ebi.ac.uk/files/${afdbId}-model_v1.cif`,
```
jyothishnt, line 127: Seen this url being used at different places. Should we move it to to global space?

jyothishnt: like you did using featureTypeToUrlPatternMap

widgets/htdocs/alphafold/exonsControlPanel.js
azangru: This is the templating syntax of the library that I am using 😃 Remember, it's not JSX and React; it is lit-html and Lit. Here's the [relevant part of documentation](https://lit.dev/docs/templates/expressions/#event-listener-expressions).

widgets/htdocs/alphafold/index.js
azangru: Yes, good point! I didn't realise that both the assets and molecule files are hosted on the same site. Updated.

widgets/htdocs/alphafold/exonsControlPanel.js
jyothishnt: ok cool. I was getting confused with both the syntaxes. Thanks

widgets/htdocs/alphafold/dataFetchers.js
```diff
@@ -0,0 +1,145 @@
...
+/**
+ * @param {Object} params - Function parameters.
+ * @param {string} params.rootUrl - Root url of the api.
+ * @param {string} params.enspId - Ensembl identifier of the protein molecule.
+ */
+export const fetchExons = async (params) => {
+  const { rootUrl: apiRoot, enspId } = params;
+  const url = `${apiRoot}/overlap/translation/${enspId}?feature=translation_exon`;
+  const exons = await fetch(url, commonOptions)
+    .then(response => response.json());
```
jyothishnt, line 36: have you checked what happens when the response is empty?

jyothishnt: We do not know if there is going to be an empty response that is going to break the code below. Better if we write this code bit defensive..  As you mentioned @azangru  what if there are genes without protein family or variation data?

azangru: I think the best we can do is check whether there is data for variants and for protein families before rendering the respective panels on the right-hand side. `dataFetchers.js` should be fine; if there's an empty array in the response, the fetchers will just return an empty array.

azangru: I added a check that makes sure that the fetched data for variants and protein features actually has some items in it.

widgets/modules/EnsEMBL/Web/Apache/Static.pm
ens-ma7: we want the function to return 1 if it matches the regex, i have been trying to get this working with one line but struggling with it. Normally we can use grep for iterating through an array but in this scenario we are doing the other way round. Maybe the foreach will be cleaner. Also i dont think `any` is a perl syntax, i am confused as how this works.

widgets/htdocs/components/95_AFDB.js
```diff
@@ -0,0 +1,95 @@
...
+
+Ensembl.Panel.AFDB = Ensembl.Panel.Content.extend({
+
+  init: function() {
+    this.base.apply(this, arguments);
+
+    var ensemblAlphafoldElement = document.querySelector('ensembl-alphafold-viewer'); // <-- custom element will be included in server response
+
+    if (!ensemblAlphafoldElement) {
+      this.onMissingData();
```
ens-ma7, line 27: The check in the `if` statement is about the presence of the element but the function `this.onMissingData()` is about "No data available". is that the right message to put here  if the widget doesnt show?

azangru: You will see [here](https://github.com/Ensembl/public-plugins/blob/1c9cff5342ba6dfa018dab2dd294c01f3e01aaef/widgets/modules/EnsEMBL/Web/Component/Transcript/AFDB.pm#L40) that the alphafold element will not be present on the page if there is no translation_object associated with this feature. This is a case of missing data.

widgets/modules/EnsEMBL/Web/Apache/Static.pm
azangru: It uses a List util that is present in our installation (https://github.com/Ensembl/public-plugins/blob/7f9fbb0a27df59b6c6aae725d11449c12f13d89c/widgets/modules/EnsEMBL/Web/Apache/Static.pm#L5). It's fine, and cleaner than the foreach.

[docs](https://perldoc.perl.org/List::Util#any)

ens-ma7: well learn a new library now and i was always using grep :) it does what we want. i am happy with this.

widgets/htdocs/components/95_AFDB.js
ens-ma7: i dont think the code will reach here if the data isn't present from the perl side. If you look at the function (the one you linked to)  it return if it doesnt find any data which means the panel below wont even be loaded which contains the js file for this. I dont think you need that check here (jsut do this.loadScript). if you want to put the check it will have to be on the perl side.

widgets/htdocs/alphafold/dataFetchers.js
jyothishnt: Thanks. Looks good.

jyothishnt (approved):

widgets/htdocs/components/95_AFDB.js
ens-ma7: @azangru not sure if you seen my comment above.

azangru: Ok, I checked what will happen if somebody tries to load alphafold view for a non-coding transcript (which wouldn't make sense from the data perspective). Turns out, perl code can already handle this just fine. Example: http://ves-hx2-75.ebi.ac.uk:8450/Arabidopsis_thaliana/Transcript/AFDB?db=core;g=AT1G04017;r=1:89235-108903;t=AT1G04017.1

I'm removing this check from the javascript side and assuming that if 95_AFDB.js is executed, it means that the `ensembl-alphafold-viewer` component is always present.

ens-st3: The availability call fails too (ie the LH menu is greyed out)

ens-ma7 (approved):

PR #486, opened 2021-10-12
Check if we have ID mapping before submitting form

ens-ap5:
As reported in this ticket: https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6242

Some bacteria lack ID mapping, which results in the user submitted jobs that inevitably fail, with a cryptic error message. Hopefully this small addition will gracefully reject jobs that don't have the data, but I haven't been able to test it in a sandbox yet. It shouldn't affect anything except the ID History Mapper.

I suggest that if it looks sensible, we merge it in and try it out on the test sites when they go up. We should ensure that it produces the error message when tried out on the bacterial species mentioned in the JIRA ticket, and also check it does not break either www or another NV site such as plants.

### DIFF
--- a/tools/modules/EnsEMBL/Web/JSONServer/Tools.pm
+++ b/tools/modules/EnsEMBL/Web/JSONServer/Tools.pm
@@ -50,6 +50,20 @@ sub json_form_submit {
     return $self->call_js_panel_method('ticketNotSubmitted', [ $sp_err ]);
   }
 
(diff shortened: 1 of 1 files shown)


ens-ap5:
As per Ridwan's suggestion, I'm making a change in the NV repo instead: https://github.com/EnsemblGenomes/eg-web-common/pull/90

PR #494, opened 2021-11-08
Rat mRatBN7.2 annotation documentation

thibauthourlier:


ens-ap5 (changes requested):
There needs to be an update to the HTML, as the file names don't match:

https://github.com/Ensembl/public-plugins/blob/postreleasefix/105/ensembl/htdocs/ssi/species/Rattus_norvegicus_annotation.html#L4

thibauthourlier:
:( I thought I properly checked the name when I created the file...

For the future, if the postreleasefix branch exists, do we still need to do a PR to master?

ens-ap5:
> :( I thought I properly checked the name when I created the file...
> 
> For the future, if the postreleasefix branch exists, do we still need to do a PR to master?

No. If there's a postreleasefix/XX branch and your commit is for release XX and any/all future releases, it goes onto that branch. If the branch doesn't exist and you want it to go into release XX, your commit needs to go onto master only.

The postreleasefix branch exists so that we can use our clever git hooks to push to master and the release branch simultaneously :)

PR #497, opened 2021-12-08
Feature/species label

ens-ap5:
Updates to bring widgets code into line with changes in ensembl-webcode.


ens-ap5:
Doh - too many extra commits!

PR #501, opened 2022-01-05
Fix VEP PDB display

ens-ap5:
Since this is specific to this tool, I think a workaround is the best option for now. It defers handling of versioned stable IDs to the API, which is (I think) the right thing to do.

See ticket https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6409


ens-ma7:
@ens-ap5 the code change looks good but i couldn't test it since there was no sandbox URL. I couldn't even see the button that was mentionned in the jira ticket on the live site. Just wanted to see how this is happening.

ens-ap5:
We did discuss this in a code review, and decided it was OK to merge it as long as we test it once the tools test is available.

PR #502, opened 2022-01-05
Add VEP benchmarking note

nakib103:
@sarahhunt @jamie-m-a 
Added notes according to benchmarking test for running VEP with variant names - 

https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-4540

### DIFF
--- a/docs/htdocs/info/docs/tools/vep/vep_formats.html
+++ b/docs/htdocs/info/docs/tools/vep/vep_formats.html
@@ -303,7 +303,16 @@ LRG_1:g.10006G>T</pre>
     the Ensembl Variation database. See <a
     href="/info/genome/variation/species/sources_documentation.html">here</a> for a list of
     identifier sources in Ensembl.</p>
-    
+   
+    <div>
+      <div class="info" style="float:left">
+        <h3>Note</h3>
+        <div class="message-pad">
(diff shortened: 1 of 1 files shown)


docs/htdocs/info/docs/tools/vep/vep_formats.html
```diff
@@ -303,7 +303,16 @@ <h2 id="id"> Variant identifiers</h2>
     the Ensembl Variation database. See <a
     href="/info/genome/variation/species/sources_documentation.html">here</a> for a list of
     identifier sources in Ensembl.</p>
-    
+   
+    <div>
+      <div class="info" style="float:left">
+        <h3>Note</h3>
+        <div class="message-pad">
+          <p>VEP is optimised for the analysis of variants in sorted VCF files or other position based formats, so input in these formats should be in preference to variant names where possible. Using variant names can take 2-5 times longer.</p>
```
jamie-m-a, line 311: Slight change:

so input in these formats should be in preference to variant names where possible
->
so input in these formats is recommended to variant names where possible

jamie-m-a (changes requested):

docs/htdocs/info/docs/tools/vep/vep_formats.html
nakib103: Updated!

jamie-m-a (approved):

jgtate (approved):

PR #503, opened 2022-01-10
Add space below each expanded transcript feature in genoverse - V2

msmanoj:
A simple solution to add spacing between the transcript features.

Only problem is that it adds some spacing above first line of feature like shown in the image below.

Screenshot 2022-01-06 at 17 41 59

Related PR:
https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6334

Sandbox URL:
http://wp-np2-1d.ebi.ac.uk:42229/Homo_sapiens/Location/View?db=core;g=ENSG00000139618;r=13:52315086-52400268;t=ENST00000544455

PR #504, opened 2022-01-10
Add space below each expanded transcript feature in genoverse

msmanoj:
An alternative solution to the following PR:
https://github.com/Ensembl/public-plugins/pull/503

It will add the required spacing without any padding at the top of the track.
Screenshot 2022-01-06 at 17 41 59 1


msmanoj:
Closing this as https://github.com/Ensembl/public-plugins/pull/503 is preferred.

PR #516, opened 2022-02-08
[ENSWEB-6362] always show GMT for tools timestamp

jyothishnt:
https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6362

Ignore BST and always show GMT for tools job timestamp


ens-ma7:
Hi @jyothishnt can we have a sandbox URL to test this please. Also based on the code, we are storing the timestamp as GMT, is that correct? Is that the default?

ens-ma7 (approved):

PR #519, opened 2022-02-15
FOR 107: add specific regulatory feature lines by type

pblins:


ens-ap5 (commented):
TBH I don't know a lot about Genoverse - I think it would be better if someone else in the web team reviewed this code.

ens-ap5:
Needs an approach that works with the changes to ensembl-webcode - see ENSWEB-6596

pblins:
Thanks, Anne :)

PR #528, opened 2022-03-31
Notify user of missing alphafold model

azangru:
When we are displaying link to alphafold model in the left-hand sidebar, we are only checking:
- that this species has some alphafold analyses
- that the transcript has a translation

We are not checking that a given transcript specifically has alphafold analysis. Which means that sometimes we show users the link that opens a page with no alphafold models available.

image

The error we are showing in such case is too non-specific and confusing for the user:

image

This PR makes the text of the error more specific for cases when the Alphafold model is missing.

Relevant Slack conversation:
https://genomes-ebi.slack.com/archives/C0JBUBFRD/p1648660444283599

Sandbox:
http://wp-np2-1e.ebi.ac.uk:8450/Glycine_max/Transcript/AFDB?db=core;g=GLYMA_12G066300;r=12:4875471-4878097;t=KRH24850

### DIFF
--- a/widgets/htdocs/alphafold/dataFetchers.js
+++ b/widgets/htdocs/alphafold/dataFetchers.js
@@ -15,8 +15,13 @@ const commonOptions = {
 export const fetchAlphaFoldId = async (params) => {
   const { rootUrl: apiRoot, enspId } = params;
   const url = `${apiRoot}/overlap/translation/${enspId}?feature=protein_feature;type=alphafold`;
-  const alphafoldFeatures = await fetch(url, commonOptions)
-    .then(response => response.json()); // should be an array of one feature
+  const alphafoldResponse = await fetch(url, commonOptions);
+  const alphafoldFeatures = await alphafoldResponse.json(); // should be an array of one feature
+
+  if (!alphafoldResponse.ok || !alphafoldFeatures.length) {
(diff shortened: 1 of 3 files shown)


jgtate (approved):

ens-st3 (approved):
Sandbox is configured to use staging-rest, which doesn't have alphafold data. I'm therefore unable to confirm that translations that do have models still work (eg http://staging.rest.ensembl.org/overlap/translation/KRH24848?feature=protein_feature;type=alphafold). Merge though if you're happy

azangru:
@ens-st3 I updated the rest hostname (locally in the sandbox, without committing) so that it now talks to the test rest api. The existing models work. Here are a couple of examples:
- [Arabidopsis transcript with an alphafold model](http://wp-np2-1e.ebi.ac.uk:8450/Arabidopsis_thaliana/Transcript/AFDB?g=AT3G52430;r=3:19431095-19434450;t=AT3G52430.1)
- [Soy transcript without an alphafold model](http://wp-np2-1e.ebi.ac.uk:8450/Glycine_max/Transcript/ProteinSummary?db=core;g=GLYMA_12G066300;r=12:4875471-4878097;t=KRH24850)

gnaamati:
Thanks, this looks great

PR #532, opened 2022-04-06
Update gene_names.html

Ben-Ensembl:
Updating gene name categories based on retirement of clone based gene names


jyothishnt:
John approved this on slack, so merging.

PR #536, opened 2022-04-13
Reg subtracks on Genoverse

msmanoj:
**NOTE**: To be merged along with https://github.com/Ensembl/ensembl-webcode/pull/882

**Description:**
- Genoverse does not seem to have the functionality to display subtracks. I have introduced a `row` key to the features data returned by `fetch_fg_regulatory_features` that tells which feature needs to go into which row. The `row` key is then used to position the features on different `y-axis` to make them appear as if they are subtracks.

- Turned off the bumping for regulation tracks

**Sandbox URL**:
http://wp-np2-1d.ebi.ac.uk:42228/Homo_sapiens/Location/View?r=8:26291508-26372680

**Related JIRA**:
https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6596

### DIFF
--- a/genoverse/htdocs/genoverse/ensembl/js/Track/library/RegulatoryFeature.js
+++ b/genoverse/htdocs/genoverse/ensembl/js/Track/library/RegulatoryFeature.js
@@ -16,15 +16,24 @@
  */
 
 Genoverse.Track.RegulatoryFeature = Genoverse.Track.extend({
-  legendType   : 'Regulation', // this forces single legend for all reg feature tracks
-  legendName   : 'Regulation Legend',
-  legend       : true,
+  legendType    : 'Regulation', // this forces single legend for all reg feature tracks
+  legendName    : 'Regulation Legend',
+  legend        : true,
+  height        : 50,
+  featureHeight : 12,
 
   view: Genoverse.Track.View.extend({
-    bump   : true,
+    bump   : false,
     labels : false,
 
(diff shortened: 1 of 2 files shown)


ens-st3:
The organisation of sub tracks is what I'd expect, but can we run this by @pblins or @ilsley to confirm ?
Compared to staging we have 'transcription factor binding sites' rather than 'Unclassified' features, and case differences (enhancer vs Enhancer)
Compared to www we've lost 'Promoter Flanking'; this may well be deliberate
Screenshot 2022-04-13 at 11 59 09
Screenshot 2022-04-13 at 11 59 02
Screenshot 2022-04-13 at 11 59 39

pblins:
Hi @ens-st3 and @msmanoj,
Thanks for the changes and comments.

- The "Unclassified" features in staging are happening because the "Transcription Factor Binding Site" type was renamed in `ensembl-webcode/conf/ini-files/COLOUR.ini` to "Transcription Factor Binding";
- The lowercases are expected as well. All legends are now lowercase;
- And losing "Promoter Flanking" also is expected, since these features are classified as "Enhancers" now.

From my side, Genoverse looks fine.
However, on bottom view, the "promoter flank" legend is a bit strange:

Screenshot 2022-04-13 at 12 32 32

Thanks.

ens-st3 (approved):

ens-ap5 (approved):
Looks like a good compromise, since we don't want to go hacking Genoverse itself!

PR #545, opened 2022-05-04
feat: Add new AF options

diegomscoelho:
## Description

Newer version of VEP cache should have two new options `--af_gnomade` and `--af_gnomadg`, also `--af_esp` should be deprecated.

## Test

...


jamie-m-a (approved):
Looks ok to me

PR #549, opened 2022-05-05
Update Atlantic salmon information (e107)

nuno-agostinho:
Based on new remapped VCF for _Salmo salar_


nakib103 (approved):
Checked in sandbox - working correctly.

PR #553, opened 2022-05-23
Improve error messages from tools

ens-ap5:
The current code for some error emails can lead to meaningless messages! I've made the code itself a bit more readable (Harpreet did love to use obscure syntax!) and added the following to the autogenerated message:

* The site name (e.g. useast.ensembl.org) should now appear in the subject line, same as for other tools exceptions
* The message body should include the HTTP referer, which will give some clues as to where to look for the problem.


ens-ma7 (approved):
Code change looks good, will be good to make sure its all displaying fine probably someone who has some previously failed jobs already on their sandbox URL or even when it goes live to just check on the live site for someone who has failed jobs in their browser session.

PR #569, opened 2022-06-30
Prairie vole EVA variants support (108)

nuno-agostinho:
Add support for remote EVA variants from Prairie vole (_Microtus ochrogaster_).

Tested in my sandbox: http://wp-np2-11.ebi.ac.uk:6070/Microtus_ochrogaster/Info/Index


jamie-m-a (dismissed):
Tested on sandbox, looks good.

PR #587, opened 2022-08-02
Add assembly converter chain files to configpacker

jyothishnt:
https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6641

This PR adds assembly converter chain files config to Configpacker.
This way there is less hassle during the release - updating configs in this ticket https://www.ebi.ac.uk/panda/jira/browse/ENSWEBREL-892

This is where we use the newly added configs will be used https://github.com/Ensembl/public-plugins/blob/release/107/tools/modules/EnsEMBL/Web/Object/AssemblyConverter.pm#L100

We may also delete the script to dump these configs mentioned in the above ticket.

http://wp-np2-1e.ebi.ac.uk:9002/

### DIFF
--- a/tools/modules/EnsEMBL/Web/ConfigPacker.pm
+++ b/tools/modules/EnsEMBL/Web/ConfigPacker.pm
@@ -164,8 +164,6 @@ sub _configure_assembly_converter_files {
     my @files = ();
     # Get all the available chain files to add entry to corresponding species
     find(sub {
-      my $file = $_;
-      my $dir  = [split /\//, $File::Find::dir]->[-1];
       push @files, s/.chain.gz$//r if /.chain.gz$/;
     }, $chain_file_dir);
 
--- a/tools/modules/EnsEMBL/Web/ConfigPacker.pm
+++ b/tools/modules/EnsEMBL/Web/ConfigPacker.pm
@@ -21,6 +21,7 @@ package EnsEMBL::Web::ConfigPacker;
 
 use strict;
 use warnings;
+use File::Find;
 
 use EnsEMBL::Web::Utils::FileHandler qw(file_get_contents);
 
@@ -30,6 +31,7 @@ sub munge_config_tree {
   my $self = shift;
   $self->PREV::munge_config_tree(@_);
   $self->_configure_blast;
+  $self->_configure_assembly_converter_files;
 }
 
 sub munge_config_tree_multi {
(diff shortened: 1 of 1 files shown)


ens-ma7:
@jyothishnt can we have a sandox to test this please. You will need to rebuild your config.packed file.

tools/modules/EnsEMBL/Web/ConfigPacker.pm
```diff
@@ -154,4 +156,22 @@ sub _vep_config_warning {
   warn $message." - thrown by ".__FILE__."\n";
 }
 
+sub _configure_assembly_converter_files {
+  my $self = shift;
+  my $species = $self->species;
+  my $version = $SiteDefs::ENSEMBL_VERSION;
```
ens-ma7, line 162: doesn't look like we are using `$version` elsewhere? do we need it? I would have thought we need the release version for the path but not sure if `ENSEMBL_CHAIN_FILE_DIR` already has it

ens-ap5: Yes it does: 

$SiteDefs::ENSEMBL_CHAIN_FILE_DIR           = defer { sprintf '/net/isilonP/public/ro/ensweb-data/%stools/%s/%s/assembly_converter', $SiteDefs::USE_SNAPSHOT_ENSWEBDATA ? 'latest/' : '', $SiteDefs::SUBDOMAIN_DIR, $release };

I think we can ommit line 162.

ens-ap5 (approved):
If Ridwan is able to confirm that Assembly Converter is getting the correct files, the code looks good to me. I was concerned about having to rebuild packed files, but I think only munge_rest() and munge_databases() write to the packed file.

https://github.com/Ensembl/ensembl-webcode/blob/main/modules/EnsEMBL/Web/SpeciesDefs.pm#L1080

tools/modules/EnsEMBL/Web/ConfigPacker.pm
```diff
@@ -154,4 +156,22 @@ sub _vep_config_warning {
...
+  my $chain_file_dir = File::Spec->catfile($SiteDefs::ENSEMBL_CHAIN_FILE_DIR, $species);
+  if (-d $chain_file_dir) {
```
ens-ma7, line 164: does that mean if a species doesnt have assembly converter files, it wont have a directory. there wont be a case where it will be an empty directory (specially if there was one in previous release and in the new release only the files have been deleted leaving an empty directory? )

jyothishnt: I believe there will be no empty dirs. I can confirm this with Production.

jyothishnt: Confirmed with production, no empty dirs. https://genomes-ebi.slack.com/archives/C0F4FQPHN/p1659604865781979?thread_ts=1659603653.602389&cid=C0F4FQPHN

tools/modules/EnsEMBL/Web/ConfigPacker.pm
```diff
@@ -154,4 +156,21 @@ sub _vep_config_warning {
...
+sub _configure_assembly_converter_files {
+  my $self = shift;
+  my $species = $self->species;
+  my $chain_file_dir = File::Spec->catfile($SiteDefs::ENSEMBL_CHAIN_FILE_DIR, $species);
+  if (-d $chain_file_dir) {
+    my @files = ();
+    # Get all the available chain files to add entry to corresponding species
+    find(sub {
+      my $file = $_;
+      my $dir  = [split /\//, $File::Find::dir]->[-1];
```
ens-ma7, line 168: what is this line doing? is $dir being used?

jyothishnt: you are right. This is not being used. Removed now. thanks

ens-ma7 (approved):

PR #600, opened 2022-08-24
Update ancestral allele file (108)

nuno-agostinho:
[ENSVAR-4827](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-4827): Renamed 107 ancestral allele file as 108


jamie-m-a (approved):
Looks good!

PR #614, opened 2022-09-16
Add Singularity instructions and improve Docker info

nuno-agostinho:
[ENSVAR-4874](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-4874)

- Check the changes in [my sandbox](http://wp-np2-11.ebi.ac.uk:6070/info/docs/tools/vep/script/vep_download.html#singularity).
- Check if the Singularity commands make sense.


msmanoj:
@nuno-agostinho  - could you confirm is this is for 108 or 109?

nuno-agostinho:
Hi @msmanoj! It would be nice to include in 108 if possible, but it's not urgent. @nakib103 still needs to review if the Singularity documentation makes sense.

nakib103 (approved):
Overall doc looks good, makes sense and seems sufficient.
LOTM! ~

PR #619, opened 2022-09-30
Show appropriate error text when Alphafold api doesn't have a model file

azangru:
This PR is for showing the appropriate message ("There is no Alphafold model for this molecule" rather than "An error occurred while loading the 3D protein viewer") if the alphafold api can't provide any data for a given id that we are sending.

Sandbox example:
http://wp-np2-1e.ebi.ac.uk:8410//Homo_sapiens/Transcript/AFDB?db=core;g=ENSG00000139618;r=13:32315086-32400268;t=ENST00000380152

_(As a sidenote, the reason this is even happening is because our rest api reports Alphafold analyses for large proteins that have been split into "fragments" (see [example query to our rest server](https://rest.ensembl.org/overlap/translation/ENSP00000369497?feature=protein_feature;type=alphafold))), whereas Alphafold is currently hiding them and [pretends](https://alphafold.ebi.ac.uk/api/prediction/P51587) that there are no models available for them. Which is why we are querying the Alphafold api with an id for which we think data exists, but the Alphafold api thinks otherwise)_

See [this Slack discussion](https://genomes-ebi.slack.com/archives/C0MUF0C3Y/p1664529478819519) for more context.

### DIFF
--- a/widgets/htdocs/alphafold/controllers/molstarController.js
+++ b/widgets/htdocs/alphafold/controllers/molstarController.js
@@ -1,4 +1,5 @@
 import { getRGBFromHex } from '../colorHelpers.js';
+import { MissingAlphafoldModelError } from '../dataFetchers.js';
 
 export class MolstarController {
 
@@ -32,7 +33,13 @@ export class MolstarController {
     const uniprotId = alphafoldId.match(parsingRegex)[1];
     const url = `${alphafoldPredictionEndpoint}/${uniprotId}`;
 
-    const alphafoldPredictionEntries = await fetch(url).then(response => response.json());
+    const alphafoldPredictionEntriesResponse = await fetch(url);
+    if (alphafoldPredictionEntriesResponse.status === 404) {
+      // no alphafold model files found
+      throw new MissingAlphafoldModelError();
+    } else if (!alphafoldPredictionEntriesResponse.ok) {
+      throw new Error('Something wrong with Alphafold prediction api');
+    }
 
     // alphafold's api will respond with an array of entries; we are interested in the first one
     const alphafoldEntry = alphafoldPredictionEntries[0];


widgets/htdocs/alphafold/controllers/molstarController.js
```diff
@@ -32,7 +33,13 @@ export class MolstarController {
     const uniprotId = alphafoldId.match(parsingRegex)[1];
     const url = `${alphafoldPredictionEndpoint}/${uniprotId}`;
 
-    const alphafoldPredictionEntries = await fetch(url).then(response => response.json());
+    const alphafoldPredictionEntriesResponse = await fetch(url);
+    if (alphafoldPredictionEntriesResponse.status === 404) {
```
ens-ma7, line 37: the status code 404 is wrong, isn't it? we might get a 404 when a request is broken or wrong endpoint or anything else we entered in the URL is wrong, it does mean we have no data but the URL is wrong. Is that confirmed from uniprot that they will send a status code 404 when there is no data found. 

To me it feels like we should check the response and if it is empty ({}) then we show this error message.

azangru: > Is that confirmed from uniprot that they will send a status code 404 when there is no data found.

It's Alphafold, not Uniprot; and yes, they seem to have a well-behaved api in this respect:

image

image

I think we can trust them on this.

ens-ma7: But this URL also give the same status code 404 but the response is different. It looks like they are sending same status code (404) but the response is different. The question again whether its better to check the response or just the status code?

https://alphafold.ebi.ac.uk/api/predictionsdfsd/P51587


image


image

azangru: > But this URL also give the same status code 404

`https://alphafold.ebi.ac.uk/api/predictionsdfsd/P51587` is a url generated according to a broken schema. The schema according to which we are querying the alphafold api is defined on lines 31-34 above. If the schema is wrong, then the alphafold widget will stop working altogether, for all molecules; and we will quickly be notified by annoyed users.

ens-ma7: Based on @azangru , we are relying on the alphafold to send us the 404 status code when there is no data and we can rely on this for being stable. That is instead of checking the response is empty, we are happy to just check the status code 404. My concern is if the URL is broken or alphafold send us the 404 status code for something else then we might have to revise this approach to check the response. 

For now the solution is working fine and i am happy with the changes.

ens-ma7 (approved):

tuanebi (approved):
I think we can assume 404 from that API means no model. So the changes look good.

PR #621, opened 2022-10-04
Update Polyphen_SIFT plugin url

nakib103:
[ENSVAR-3718](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-3718)

**This can be for e109**


nuno-agostinho (approved):
Seems good to me!

PR #625, opened 2022-10-13
Variant Recoder fix HGMD url

dglemos:
Add URL to all HGMD variants (release 109).

Sandbox: http://wp-np2-11.ebi.ac.uk:7040/Homo_sapiens/Tools/VR
Variant to test: rs138742754
HGMD in output is `BM1115006`


jyothishnt:
http://wp-np2-11..ebi.ac.uk:7040/Homo_sapiens/Tools/VR/Ticket?tl=5ZXGneQ8hzV2rWui

HGMD output is linking to the homepage http://www.hgmd.cf.ac.uk/ac/index.php 
without any ID. Should this be linking to the page for `BM1115006`?

dglemos:
We will let it be as it is for now.

PR #632, opened 2022-11-11
Update VEP Docker docs (e109)

nuno-agostinho:
[ENSVAR-2498](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-2498): avoid showing the (now) unneeded "data structure" for the local folder with VEP data to avoid confusing users as in https://github.com/Ensembl/ensembl-vep/issues/616

Also, update information on using VEP Docker/Singularity images based on updates performed for e109.

Check the changes to the help page in my sandbox: http://wp-np2-11.ebi.ac.uk:6070/info/docs/tools/vep/script/vep_download.html#docker


likhitha-surapaneni (approved):
Looks good to me.

PR #639, opened 2022-12-09
Update VEP plugins docs (109)

nuno-agostinho:
* Mention support for VEP plugins in Docker image
* Mention that some plugins are available in REST
* Remove mention to external UTRannotator plugin (we now include it in VEP_plugins repo)

The changes are reflected in my [sandbox](http://wp-np2-11.ebi.ac.uk:6070/info/docs/tools/vep/script/vep_plugins.html).


nakib103 (approved):
Looks Good! ~

PR #641, opened 2022-12-13
feat: Public docs update (e109) - Variation

diegomscoelho:
## Description

Changes in variation docs. The changes can be seen in the sandbox - http://wp-np2-11.ebi.ac.uk:7080/

## Test

Go to .html files that are different and check if it's OK.


nakib103 (approved):
Looks good! ~ thanks Diego.

PR #644, opened 2022-12-19
feat: New info about HGVS_OFFSET behavior (e109)

diegomscoelho:
## Description

Alteration to inform HGVS_OFFSET value behavior. Regarding issue (https://github.com/Ensembl/ensembl-vep/issues/1328)

## Test

Just check new text information.
http://wp-np2-11.ebi.ac.uk:7080/info/docs/tools/vep/vep_formats.html#vcfout
http://wp-np2-11.ebi.ac.uk:7080/info/docs/tools/vep/script/vep_options.html


jamie-m-a:
Prefer text in vep_formats.html to be: "Value must be greater than 0."

Otherwise looks good.

jamie-m-a (changes requested):
Text changes as per comments

PR #652, opened 2023-01-20
Update Variation consequences (e110)

nuno-agostinho:
* Convert consequences image from JPG to SVG
* Add [`coding_transcript_variant consequence`](http://www.sequenceontology.org/miso/current_svn/term/SO:0001968), a recently supported consequence: https://github.com/Ensembl/ensembl-variation/pull/947

Sandbox: http://wp-np2-11.ebi.ac.uk:6070/info/genome/variation/prediction/predicted_data.html


diegomscoelho (approved):
LGTM! Checked on Nuno's sandbox and all look fine.

PR #605, opened 2022-08-31
Update Variant Recoder

dglemos:
Add link to HGMD: [ENSVAR-4363](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-4363)
MANE select only available for human: [ENSVAR-4364](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-4364)

Sandbox: http://wp-np2-11.ebi.ac.uk:7040/Homo_sapiens/Tools/VR

PR #655, opened 2023-02-01
Display MANE select option only for human in Variant Recorder (duplicate)

imransl:
MANE select option in Variant Recorder was being displayed for all species, despite it only being available for human in the data. This was fixed in the code in both Perl and JavaScript side.

Related PR: #605 
Sandbox link: http://wp-np2-1e.ebi.ac.uk:8310/Homo_sapiens/Tools/VR

### DIFF
--- a/tools/modules/EnsEMBL/Web/Component/Tools/VR/InputForm.pm
+++ b/tools/modules/EnsEMBL/Web/Component/Tools/VR/InputForm.pm
@@ -221,19 +221,18 @@ sub get_cacheable_form_node {
     }]
   });
 
-  if ($current_species eq 'Homo_sapiens') {
-    $input_fieldset->add_field({
+  # Add Mane Select option separately with a different classname for ease of toggling
+  $input_fieldset->add_field({
     'type'        => 'checklist',
-    'field_class' => [qw(_stt_yes _stt_allele _stt_Homo_sapiens)],
+    'field_class' => [qw(_stt_yes _stt_allele _stt_mane_select)],
     'name'        => 'mane_select',
     'values'      => [{
       'caption'     => $fd->{mane_select}->{label},
       'helptip'     => $fd->{mane_select}->{helptip},
       'value'       => 'yes',
       'checked'     => 0
-      }]
-    });
-  }
+    }]
+  });
 
   # Run button
   $self->add_buttons_fieldset($form);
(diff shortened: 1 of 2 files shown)


tools/modules/EnsEMBL/Web/Component/Tools/VR/InputForm.pm
```diff
@@ -221,19 +221,18 @@ sub get_cacheable_form_node {
     }]
   });
 
-  if ($current_species eq 'Homo_sapiens') {
-    $input_fieldset->add_field({
+  # Add Mane Select option separately with a different classname for ease of toggling
+  $input_fieldset->add_field({
     'type'        => 'checklist',
-    'field_class' => [qw(_stt_yes _stt_allele _stt_Homo_sapiens)],
```
jyothishnt, line 227: Wonder why this was not working in the first instance. As far as I know this is how we enable a setting for a particular species. Wasn't this working before?

imransl: @jyothishnt Are you referring to the `field_class` change? It really had nothing to do with why it didn't work, but just thought renaming the `_stt_Homo_sapiens` to `_stt_mane_select` would make it more self descriptive.

jyothishnt: really? lets discuss tomorrow

jyothishnt:
Notes from investigation and discussion: 
Cause of the issue is selectToToggle is broken due to https://github.com/Ensembl/ensembl-webcode/blob/release/108/htdocs/components/00_jquery_selecttotoggle.js#L32
https://github.com/Ensembl/ensembl-webcode/pull/908

But Chicken breeds or other species that have `.` in their names would still break. But that is minor and can be fixed after release 109

dglemos:
Hi @imransl, thanks for the update. I cannot select different species on your sandbox but by testing on my sandbox I can confirm the option mane is not available for other species - I tested the main species we have in variation. 
Unfortunately, we have the same issue in VEP. Options like SpliceAI should only be available for human but they are displayed for all species.

imransl:
This PR is apparently in the wrong branch. Closing this for #661.

imransl:
> Hi @imransl, thanks for the update. I cannot select different species on your sandbox but by testing on my sandbox I can confirm the option mane is not available for other species - I tested the main species we have in variation. Unfortunately, we have the same issue in VEP. Options like SpliceAI should only be available for human but they are displayed for all species.

Thanks for testing @dglemos. I have fixed these issues in two different PRs, one in public-plugins itself and the other in ensembl-webcode. I had to close this PR since it is in the wrong branch.

PR #664, opened 2023-02-17
Add Geno2MP plugin to web VEP (110)

nuno-agostinho:
[ENSVAR-2852](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-2852)

### Testing

Use `rs28775022` to test web VEP with Geno2MP enabled in my sandbox: http://wp-np2-11.ebi.ac.uk:6070/Tools/VEP

### DIFF
--- a/tools/conf/vep_plugins_web_config.txt
+++ b/tools/conf/vep_plugins_web_config.txt
@@ -35,6 +35,10 @@
     "available" => 1
   },
 
+  Geno2MP => {
+    "available" => 1
+  },
+
   GO => {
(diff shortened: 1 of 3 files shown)


azangru:
@nuno-agostinho Is it expected that the urls in the `Geno2MP URL` column are plain text rather than link tags?

image

nuno-agostinho:
Hi @azangru! The results from VEP are plain text links, so I just modified `Results.pm` to turn them into clickable URLs. Thanks for your suggestion!

PR #671, opened 2023-03-21
Add new and updated vcf-only species

nakib103:
Tested using selenium. The test suite are available in this PR - https://github.com/Ensembl/ensembl-internal-variation/pull/112


nuno-agostinho (approved):
All seem good to me: all variation examples are working. Thanks, @nakib103!

PR #674, opened 2023-03-31
Update Regulation documentation for e!110

ilsley:
We have added regulatory data for a number of species using our new pipeline. This needs to be added to the help & docs.
We have also added a page for inbound links from a popular 2013 paper.

**Sandbox URL:**
Documentation is currently available on http://wp-np2-11.ebi.ac.uk:6010/info/genome/funcgen/index.html.


ensembl/htdocs/info/genome/funcgen/index.html
```diff
@@ -95,9 +104,7 @@ <h2 id="references">References</h2>
         <em>Database</em> pii: bav119<br />
       </p>
     </li>
-  </ul>
-
-
+    </ul>
```
jgtate, line 107:
```suggestion
    </ol>
```

PR #681, opened 2023-04-19
Improve docs on VEP installation

nuno-agostinho:
* Document usage of all in species option ([ENSVAR-5527](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5527))
* Add anchor links to all documented options for INSTALL ([ENSVAR-5526](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5526))

Test in my sandbox: http://wp-np2-11.ebi.ac.uk:6070/info/docs/tools/vep/script/vep_download.html#opt_species

### DIFF
--- a/docs/htdocs/info/docs/tools/vep/script/vep_download.html
+++ b/docs/htdocs/info/docs/tools/vep/script/vep_download.html
@@ -960,7 +960,7 @@ cd ensembl-vep-release-[[SPECIESDEFS::ENSEMBL_VERSION]]/</pre>
       <th>Alternate</th>
       <th>Description</th>
     </tr>
-    <tr>
+    <tr id="opt_assembly">
       <td><pre>--ASSEMBLY</pre></td>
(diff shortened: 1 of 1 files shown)


jyothishnt:
Hi @nuno-agostinho  We can see the tr ids but no anchor tags. Is this expected?

nuno-agostinho:
Hey @jyothishnt, what do you mean with anchor tags exactly?

The idea of this update is to have `tr id` for all command-line options in the documentation so we can link users directly to those options using an URL, like so: http://wp-np2-11.ebi.ac.uk:6070/info/docs/tools/vep/script/vep_download.html#opt_species

The changes in this PR reflect what we already do in this script: https://github.com/Ensembl/public-plugins/blob/88a5a1a840f70fe7ec9af11b6f0a43854acd3db2/docs/htdocs/info/docs/tools/vep/script/vep_options.html#L132-L144

PR #682, opened 2023-05-02
Update Mastermind and SpliceAI file version (e110)

dglemos:
Update version for release 110.
SpliceAI ensembl_mane files are only available for grch38.


nakib103 (approved):
LOTM!~

PR #689, opened 2023-05-16
Update tag for replacing content for public docs (e111)

nakib103:
Related to - https://github.com/Ensembl/ensembl-variation/pull/989


nuno-agostinho (approved):
Looks good to me!

PR #694, opened 2023-05-25
Add new flag description to VEP public doc (e110)

nakib103:
related to - https://github.com/Ensembl/ensembl-vep/pull/1421

sandbox test url - http://wp-np2-11.ebi.ac.uk:7070/info/docs/tools/vep/script/vep_options.html#opt_hgvsg


nuno-agostinho (changes requested):

nuno-agostinho (approved):
Thanks for the changes, @nakib103! All seems good to me!

PR #702, opened 2023-06-20
Update VEP consequences table (110)

nuno-agostinho:
Add missing consequences to Variation table

This table is used in https://www.ensembl.org/info/genome/variation/prediction/predicted_data.html


likhitha-surapaneni (approved):
LGTM, thank you

PR #704, opened 2023-06-21
Ga4

jyothishnt:
This PR adds support for google analytics 4 which is needed as Google is due to shutdown universal analytics in July.

It was decided to add support along side universal analytics due to an unfortunate overlap of the reporting required for grants and Google shutting down analytics. Doing this de-risks generating the reports that are required by several grants.

The idea with the code changes is to make it easy to remove universal support at a later date.

More detail can be found here https://www.ebi.ac.uk/panda/jira/browse/ENSWEB-6617 including test property and deployment details.

**NOTE: Need an update in ebi-plugins domain codes before pushing this.**


ga/htdocs/components/90_ga.js
```diff
@@ -210,51 +276,191 @@ Ensembl.GA = {
...
+      s.type = "text/javascript";
+      s.src = "https://www.googletagmanager.com/gtag/js?id=" + this.code();
+      s.setAttribute('async', '');
+      $("head").append(s);
 
-Ensembl.GA.EventConfig.prototype.instantiate = function (extra, defaults, e) {
-  var config  = $.extend({}, defaults, this, extra);
-  config.data = $.extend({}, this.data);
+      // get the species regexp before calling filterURL
+      this.urlSpeciesRegex = new RegExp('/(' + Ensembl.allSpeciesList.join('|') + ')/');
```
jyothishnt, line 309: This is repeated in both GA4 and GAU. Better move it to GA where it is actually used?

ga/htdocs/components/90_ga.js
```diff
@@ -14,20 +14,16 @@
...
      * List of configs registered by default (added later in ga_configs.js)
      */
-
-  domainCodes: {},
-    /*
-     * key pair value for domain name and it's corresponding code (specified later in plugins)
-     */
-
-  verbose: false,
+  verbose: true,
```
jyothishnt, line 26: Do you need this to be turned of always?

PR #708, opened 2023-06-29
GA blast fix

jonk-ebi:
This PR corrects a couple of errors introduced during integrating GA4

Currently deployed to http://wp-np2-1f.ebi.ac.uk:10000/

Currently connected to https://analytics.google.com -> Dev GA4

### DIFF
--- a/ga/htdocs/components/90_ga.js
+++ b/ga/htdocs/components/90_ga.js
@@ -61,15 +61,19 @@
    },
    
    getConfig: function (configId) {
-    if(Ensembl.GA4.initialised)
-    {
-      Ensembl.GA4.getConfig(configId);
+    /*
+     * Grabs a config from the eventConfigs array with the given id
+     */
+    if (Ensembl.GA4.initialised === true || Ensembl.GAU.initialised === true) {
+      throw "Can not get hold of a config after Ensembl.GA4 has been initialised";
     }
-    else
-    {
-      Ensembl.GAU.getConfig(configId);
+
+    for (var i = 0; i < Ensembl.GA.eventConfigs.length; i++) {
+      if (Ensembl.GA.eventConfigs[i].id && Ensembl.GA.eventConfigs[i].id === configId) {
+        return Ensembl.GA.eventConfigs[i];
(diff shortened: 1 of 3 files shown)


azangru:
I've only checked the sandbox. It does not show an error on the BLAST results page; but on the other hand, I remember that the error that Tuan noticed was coming out of the ga_config_tools.js file, where `Ensembl.GA.getConfig` was undefined. This time I looked at this file in the debugger, the `Ensembl.GA` itself was undefined, which meant that [this condition](https://github.com/Ensembl/public-plugins/blob/release/109/tools/htdocs/components/95_ga_config_tools.js#L22) was falsy, and that the line that was giving the error on rapid didn't even run:

image

I don't know what this all means, or whether this is expected 🙂

jonk-ebi:
> I've only checked the sandbox. It does not show an error on the BLAST results page; but on the other hand, I remember that the error that Tuan noticed was coming out of the ga_config_tools.js file, where `Ensembl.GA.getConfig` was undefined. This time I looked at this file in the debugger, the `Ensembl.GA` itself was undefined, which meant that [this condition](https://github.com/Ensembl/public-plugins/blob/release/109/tools/htdocs/components/95_ga_config_tools.js#L22) was falsy, and that the line that was giving the error on rapid didn't even run:
> 
> image
> 
> I don't know what this all means, nor whether this is expected 🙂

Appears to be an artefact of the debug mode, 90_ga.js is pending on the network when 95 executes. 

I can confirm that the event `LocalContext-LeftButton` is triggered on the sandbox when clicking the left hand buttons such as configure this page. 

`selector        : '.local_context a, .tool_buttons a',`
Screenshot 2023-06-30 at 10 28 55

azangru:
I swear it doesn't run for me even in a minified file (recording attached); but after all, the important thing is it doesn't break the BLAST results page. 

:shipit: 


https://github.com/Ensembl/public-plugins/assets/6834224/0b9415bc-bc3d-4bae-92be-01206fb6b894

azangru (approved):

PR #709, opened 2023-06-29
Update VEP Input Form docs with all available plugins

nuno-agostinho:
Some web VEP plugins were not documented in VEP Input Form. This PR adds the missing docs for those plugins.

Test in my sandbox: http://wp-np2-11.ebi.ac.uk:6070/info/docs/tools/vep/online/input.html


dglemos (approved):

jyothishnt:
Is this intended for 110 or 111? @nuno-agostinho @dglemos

nuno-agostinho:
Hey @jyothishnt, we answered at the time to @jgtate to merge if possible for 110, but it's okay if it only goes to 111 as it is not urgent to fix. Thanks!

PR #717, opened 2023-08-15
Update web VEP input detection (e111)

nuno-agostinho:


tools/htdocs/components/20_VEPForm.js
```diff
@@ -261,14 +266,30 @@ Ensembl.Panel.VEPForm = Ensembl.Panel.ToolsForm.extend({
    */
     var data = input.split(/\s+/);
 
-    // SPDI: 1:230710044:A:G
+    // region: chr21:10-10:1/A
     if (
+      data.length === 1 &&
+      data[0].match(/^([^:]+):(\d+)-(\d+)(:[-\+]?1)?[\/:]([a-z]{3,}|[ACGTN-]+)$/i)
```
jyothishnt, line 272: This regex looks complex.
1. Can you please break it down and also move it to a separate method? 
2. If you are marking it case insensitive you don't need to capitalise ACGTN.  
3. Please add in more examples of regions in the comments (line 269)

nuno-agostinho: Hey @jyothishnt, when you mention breaking it down, do you mean something like this?

```js

var chrom  = "([^:]+)";
var pos    = "(\\d+)-(\\d+)";
var strand = "(:[-\+]?1)?";
var allele = "([a-z]{3,}|[ACGTN-]+)";

const re   = new RegExp(`^${chrom}:${pos}${strand}[\/:]${allele}$`, "i");
data[0].match(re);
```

azangru: > return;
> ...
>  return 'region';

Sorry, I meant return false or return true in my snippet above. I imagined it as a method that would be called inside of the if-condition that currently contains the long regex. Something like `if (this.isRegionFormat(data)) { return 'region'; }`.

nakib103: Hi @jyothishnt @azangru ,

Thanks for reviewing this PR.

- Breaking that function by splitting `:` has a problem because the elements are not always separated by `:`.

- This function implements the same logic as `detect_format` function used in ensembl-vep and ensembl-variation. Having same logic in different places was causing issue when we are doing updates in one places and the other place was not being updated. So, we recently made changes to have to only have one function in ensembl-variation - 
https://github.com/Ensembl/ensembl-vep/pull/1477
https://github.com/Ensembl/ensembl-variation/pull/1026
Unfortunately we need to keep this function also (making changes in 2 places is better than 3 at least !). Saying all these things is prelude to saying it is better for us if we can keep the logic same as the other function so it is easy to update in future - 
https://github.com/Ensembl/ensembl-variation/blob/86bad9e537344c88df5fb04e8ee0fce0a7d59876/modules/Bio/EnsEMBL/Variation/Utils/VEP.pm#L388

Let me know if that makes sense.

azangru: > Breaking that function by splitting : has a problem because the elements are not always separated by :.

But your regular expression was also checking for the presence of `:`, wasn't it? If a string does not have a `:`, it would fail this particular check (`parts.length` will not equal 3), just as the regex would.

Anyway, if you have a strong preference for the long regex, this is fine by me. It is not the new site, after all.

nakib103: This part of the regex allows the second `:` to be not present - 
`(:[-\+]?1)?[\/:]`

Example here - 
https://www.ensembl.org/info/docs/tools/vep/vep_formats.html#region

Yeah, it is better to not introduce a potential breaking changes for the current sites. Thanks!

nakib103 (approved):

PR #718, opened 2023-08-23
Fix a typo in vep_faq.html

amushtaq102:
in the section: Why do I see multiple known variants mapped to my input variant?

I have changed: "VEP compares (you) > (your) input to known variants from the Ensembl variation database. In some cases one input variant can match multiple known variants:


azangru:
Note that the base branch needs fixing.

PR #719, opened 2023-09-04
Remove non-working vep example from INI file (e111)

nakib103:
[ENSVAR-5904](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5904)


nuno-agostinho (approved):
All seems good, thanks @nakib103!

nuno-agostinho (approved):

PR #712, opened 2023-07-12
Web VEP: add Open Targets Genetics option (e111)

nuno-agostinho:
[ENSVAR-5840](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-5840): Open Targets plugin
[ENSVAR-2120](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-2120): Return reference and uploaded alleles (hidden by default)

* Requires https://github.com/Ensembl/ensembl-webcode/pull/975
* Includes changes from https://github.com/Ensembl/public-plugins/pull/709

## Changelog

* [Add OpenTargets option to web VEP](https://github.com/Ensembl/public-plugins/commit/90c93a27debea35b27282302bce17a2f882b572f)
* [Format web VEP results for OpenTargets plugin](https://github.com/Ensembl/public-plugins/commit/933622b5c7336bc4e5fef36ddec1e30c459404d9) (+ [improvements](plugins/pull/712/commits/ca6d75a6cbbcaa6eae1472bbd68fd1074ccf852a))
* [Document Open Targets Genetics option](https://github.com/Ensembl/public-plugins/commit/1c73482a108eaf43d2e9c7fcc7fd74729a5f0963)

### Other changes

* [Return reference allele column (hidden by default)](https://github.com/Ensembl/public-plugins/commit/674ef4b3fd137da64920de87fd6f642c41e2708f), required to create the external URL to Open Targets
* [Return uploaded allele column (hidden by default)](https://github.com/Ensembl/public-plugins/pull/712/commits/898b20c721f9af71d804479a44acf8c86e36349b)
* [Simplify Geno2MP code](https://github.com/Ensembl/public-plugins/commit/3e340c21c956bc6b7be836b1eaaae5f64cfcbcc4)
* Improve public documentation on Input Form:
    - [Use HTML forms instead of screenshots](https://github.com/Ensembl/public-plugins/pull/712/commits/f70385303370972f7b7b622bf6161e643aa646e4)
    - [Improve Jobs section](https://github.com/Ensembl/public-plugins/pull/712/commits/80f1f5de2799b3ca4161c3fb5516c460b03baac9)

## Testing

1. Check documentation changes in http://wp-np2-11.ebi.ac.uk:6070/info/docs/tools/vep/online/input.html
2. Run web VEP using Geno2MP and Open Targets Genetics options enabled in http://wp-np2-11.ebi.ac.uk:6070/Tools/VEP
[...]

PR #720, opened 2023-09-06
Web VEP: remove redundant 'Variant data' section code (e111)

nuno-agostinho:
After merging #712 and #716, the web VEP input form ended up with redundant code for the 'Variant data' section.

This PR intends to remove the redundant code.

### Testing

Check results in my sandbox: http://wp-np2-11.ebi.ac.uk:6070/Multi/Tools/VEP?db=core


nakib103 (approved):
LOTM!~

PR #728, opened 2023-09-20
Document `--custom` usage for older versions of VEP (e111)

nuno-agostinho:
Fixes Ensembl/ensembl-vep#1453

## Description

In VEP 110, we improved `--custom` to support a more flexible approach with key-value pairs. This change is incompatible with VEP 109 and earlier.

In this PR, we add the usage documentation for these older versions of VEP so they are still easily available to all users, as requested by a user in Ensembl/ensembl-vep#1453.

## Changelog

- Re-introduce 109 docs for `--custom` usage
  - This section is virtually the same from 109 docs: https://github.com/Ensembl/public-plugins/blob/release/109/docs/htdocs/info/docs/tools/vep/script/vep_custom.html#L139-L235.
  - The usage of positional arguments (like in older VEP versions) is still compatible with VEP 110 and newer.
- Update the section regarding the modern usage of `--custom` to:
  - emphasise the advantages of using key-value pairs
  - explicitly state that `file` and `format` are required
  - improve description and accepted values for some options
  - [remove wrong footnote for `Force report coordinates`](https://github.com/Ensembl/public-plugins/pull/728/commits/3a1fce9cf046b670beabde955bb361cd6b37d19f) (pointed by @nakib103)

## Testing

Check changes in my sandbox: http://wp-np2-11.ebi.ac.uk:6070/info/docs/tools/vep/script/vep_custom.html#custom_options

### DIFF
--- a/docs/htdocs/info/docs/tools/vep/script/vep_custom.html
+++ b/docs/htdocs/info/docs/tools/vep/script/vep_custom.html
@@ -381,7 +381,7 @@ tabix -p bed myData.bed.gz</pre>
       <td><pre>coords</pre></td>
       <td><kbd>0</kbd> (default) or <kbd>1</kbd></td>
       <td>
-        <b>Force report coordinates (*):</b>
+        <b>Force report coordinates:</b>
         <ul>
           <li>
             <kbd>0</kbd>: outputs the identifier field (or value in the case of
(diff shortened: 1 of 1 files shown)


dglemos (approved):
The documentation looks good!
Note, if you change the behaviour to only support bigwig don't forget to update the description of `summary_stats`.

dglemos:
This update should be included in release 111.

docs/htdocs/info/docs/tools/vep/script/vep_custom.html
```diff
@@ -138,15 +138,124 @@ <h3>Note</h3>
...
+    ./vep [...] --custom clinvar.vcf.gz,ClinVar,vcf,exact,0,CLNSIG,CLNREVSTAT,CLNDN \
+                --custom TOPMED_GRCh38_20180418.vcf.gz,topmed_20180418,vcf,exact,0,TOPMED \
+                --custom UK10K_COHORT.20160215.sites.GRCh38.vcf.gz,uk10k,vcf,exact,0,AF_ALSPAC</pre>
+
+  </div>
+
+  <br/>
+  <div class="navbar" onclick="show_hide('custom_options');" id="custom_options" style="padding-left:5px;margin-bottom:0px;cursor:pointer"><img id="img_custom_options" src="/i/16/minus-button.png" style="vertical-align:middle"/> <span style="position:absolute;left:28px;top:9px">Using key-value pairs in <kbd>--custom</kbd> with VEP [[SPECIESDEFS::ENSEMBL_VERSION]]</span></div>
+  <div id="div_custom_options" style="display:inline">  
+  </br>
```
jgtate, line 249:
```suggestion
  <br />
```

PR #733, opened 2023-10-18
Update variation data in public docs (e111)

nakib103:
sandbox: http://wp-np2-11.ebi.ac.uk:7070/info/genome/variation/index.html
For variant classification page you can check - http://wp-np2-11.ebi.ac.uk:6070/info/genome/variation/prediction/classification.html
My sandbox is not working


jamie-m-a (approved):
Look good to me - thanks!

PR #738, opened 2023-11-13
Add original author name in UTRannotator plugin (e111)

nakib103:
sandbox url for checking - 
http://wp-np2-11.ebi.ac.uk:7070/info/docs/tools/vep/script/vep_plugins.html#utrannotator


jamie-m-a (changes requested):
Looks great, thanks for updating.

Can you also change the developer column - where it currently says Ensembl to also add Xiaolei Zhang. Can use bullet points like for MTR

nakib103:
Thanks @jamie-m-a , updated!

jamie-m-a (approved):
Perfect - thanks Nakib

PR #750, opened 2023-12-08
fixed both sheep ncbi_golden_path

EreboPSilva:
Fixed both sheep ncbi_golden_path to taxon_id which seems to be the correct value.
@jyobhai


jyothishnt (approved):
Looks good to me

PR #754, opened 2024-01-17
Enable regulatory feature for pig and chicken (e111)

nakib103:
In e111 we have new regulatory build for pig and chicken. But web VEP does not show option to annotate regulatory feature. This PR enable option for regulatory region consequences in webVEP.

NB: also adding a small update to VEP plugin documentation for CADD.

Test
sandbox: http://wp-np2-11.ebi.ac.uk:7070/Gallus_gallus/Tools/VEP/Results?tl=ztzzZXLJ16krSpEU-40

(plugin doc) http://wp-np2-11.ebi.ac.uk:7070/info/docs/tools/vep/script/vep_plugins.html#cadd


jgtate:
When looking at your sandbox, @nakib103 , we can see the extra section appear for chicken, but there's no checkbox: 
image
The checkbox does appear for human though:
image
Is this because the chicken and pig packed files need to be recreated on your sandbox, or an actual bug?

We're also seeing some perl debug leaking into the top of the page, though it's possibly something enabled in the checkout but not committed: 
image

jgtate (approved):

PR #756, opened 2024-01-26
Fix wrong categories for VEP plugins in What's New (e111)

nuno-agostinho:
[ENSVAR-6215](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-6215)

Sandbox: http://wp-np2-11.ebi.ac.uk:6070/info/docs/tools/vep/script/vep_download.html#new


jamie-m-a (approved):
Looks good, thanks Nuno!

PR #761, opened 2024-02-06
Improve documentation about un-supported GTF/GFF types (e112)

nakib103:
sandbox links - 
FAQ - http://wp-np2-11.ebi.ac.uk:7070/info/docs/tools/vep/vep_faq.html
Annotation source - http://wp-np2-11.ebi.ac.uk:7070/info/docs/tools/vep/script/vep_cache.html#gff


nuno-agostinho (changes requested):
Thanks for the PR, @nakib103! Here are some suggestions.

PR #766, opened 2024-02-19
Check config tree instead of INI for regulation species (e113)

nakib103:
Currently we [update INI files](https://github.com/Ensembl/public-plugins/pull/754) to setup species with regulatory build in web VEP. But it can be directly derived from the species def config tree.

The target of this update is in e113 - regulation will review the [summary data](https://github.com/Ensembl/ensembl-webcode/blob/25c16dc3ff2135b5e3b31e0ad186113bb25194d3/modules/EnsEMBL/Web/ConfigPacker.pm#L699) loaded in the species defs first [confirmed].

test sandbox url - http://wp-np2-11.ebi.ac.uk:7070/Tools/VEP

Test can be done for `turbot`, `european seabass` on example variants.


jyothishnt:
Hi @nakib103 , any update on this?

nakib103:
Hi @jyothishnt, it is not ready yet. We are looking into it.

likhitha-surapaneni (approved):
LGTM, thanks @nakib103

PR #776, opened 2024-03-13
AlphaMissense license update

sarahhunt:
Hi @jonk-ebi - this is the docs update as discussed


jonk-ebi:
It might be worth removing the `<br />`s from the opening paragraph and the disclaimer as they can interrupt the flow on different sized screens. 

image

image

Other than that it looks good

sarahhunt:
Thanks for taking a look @jonk-ebi . These docs are automatically generated, so we'll need to alter the code to fix the breaks permanently, but I've removed them here.

jonk-ebi (approved):

PR #782, opened 2024-03-27
pdf_updates

jamie-m-a:
e112


nuno-agostinho (approved):
The PDF seems to have all the latest documentation updates.

Some links seem to direct to the sandbox, note/warning elements are not displayed and collapsible elements button do nothing when clicking on them, but these seem to be long-standing issues. Thanks, @jamie-m-a!

PR #788, opened 2024-04-23
Genebuild static for 112

EreboPSilva:
Adding/modifying 21 html files (one per annotation produced), and a collective PDF for the clade.


veidenberg (approved):
Looks good to me.

PR #792, opened 2024-05-16
Adding missing cli plugin (e112 - if possible!)

jamie-m-a:
Hi web - so sorry, was just arranging the VEP release when I realised we'd missed a plugin off the list for the what's new section.

Is it possible to add it in? It's a very tiny text change


olaaustine (approved):
LGTM

PR #796, opened 2024-05-29
Improve VEP FAQ

nuno-agostinho:
[ENSVAR-6224](https://www.ebi.ac.uk/panda/jira/browse/ENSVAR-6224):
- Avoid generic questions like "Why do I see the following error"
- Add anchor links to every question
- Use `<h3>` heading for the questions

Check in [my sandbox](http://wp-np2-11.ebi.ac.uk:6070/info/docs/tools/vep/vep_faq.html#indel-vcf).


olaaustine (approved):
LGTM! @nuno-agostinho 
Thank you

PR #803, opened 2024-07-22
Fixing miRNA docs

jamie-m-a:
The mirna flag doesn't work for refseq transcripts and no trivial way to add this functionality, so we will indicate that it's only suported for Ensembl/genocde


olaaustine (approved):
LGTM!

PR #811, opened 2024-08-02
VEP doc: add default value for --max_sv_size

nakib103:
sandbox url: http://wp-np2-11.ebi.ac.uk:7070/info/docs/tools/vep/script/vep_options.html


nuno-agostinho (approved):
Looks good to me, thanks @nakib103!

PR #829, opened 2024-09-04
Add missing perl library dependencies needed to run VEP (e114)

nakib103:
We have added the following libraries in our API code at some point is previous releases -
-[ LWP::Simple](https://github.com/Ensembl/ensembl-vep/blob/31a3581b84495b617b2f3980da6c6313ca6d238f/modules/Bio/EnsEMBL/VEP/AnnotationSourceAdaptor.pm#L73)
-[ List::MoreUtils](https://github.com/Ensembl/ensembl-vep/blob/31a3581b84495b617b2f3980da6c6313ca6d238f/modules/Bio/EnsEMBL/VEP/Parser/VCF.pm#L82)
We should add them to the dependency list in the installation instruction in the public docs


jamie-m-a (approved):
Thanks for updating!

PR #834, opened 2024-10-08
Remove sheep texel from web VEP

nakib103:
https://www.ebi.ac.uk/panda/jira/browse/ENSINT-2018

sandbox url - http://wp-np2-11.ebi.ac.uk:7070/Tools/VEP - check sheep texel not available in species list.


jyothishnt (approved):
Looks good to me.

PR #839, opened 2024-11-05
VEP docs - pangenome protein function link does not work

nakib103:
The link contains http which browser is not allowing to download. Replaced it with https.

sandbox url - http://wp-np2-11.ebi.ac.uk:7070/info/docs/tools/vep/script/vep_example.html#pangenomes

PR #840, opened 2024-11-05
VEP docs - pangenome protein function link does not work

nakib103:
https://github.com/Ensembl/public-plugins/pull/839


jamie-m-a (approved):
Looks good, thanks Nakib

PR #842, opened 2024-11-14
Tools hive: LSF to Slurm migration

azangru:
The changes in this PR are part of switching `ensembl-hive` from version 2.2 to version 2.7, and of migration from LSF to Slurm. They include:

- Updates to `beekeeper_manager.pl` to stop it from passing arbitrary command-line options to `beekeeper.pl` (beekeeper.pl in ensembl-hive, which is the script that is eventually called, no longer allows options outside of the ones that it expects).
- Updates to `beekeeper.pl` to fix a problem when it would call itself. You will notice that the old version of `beekeeper.pl` has a `do` command, with which it calls `beekeeper.pl` in `ensembl-hive`. For some reason that remains a mystery, after we switched ensembl-hive to version 2.7, our `beekeeper.pl` (i.e. the one in the public-plugins repo) started calling itself at the `do` line instead of calling `beekeeper.pl` in `ensembl-hive`. I have updated the file to use `fork` instead of `do`.
- Updates to `ToolsPipeConfig.pm` and its subclasses to update the definition of jobs for SLURM


imransl (approved):
Looks good to me

jyobhai (approved):

PR #854, opened 2025-02-12
Document multiple chromosome files feature in VEP custom annotation (e114)

nakib103:
VEP custom annotation allows to give multiple files split by chromosome using `###CHR###` value but not documented. This PR document this feature.

sandbox url - http://wp-np2-11.ebi.ac.uk:7070/info/docs/tools/vep/script/vep_custom.html


jamie-m-a (approved):
Approved

PR #866, opened 2025-04-09
Avoid hive param_substitute in VEP custom config (114)

nakib103:
We use ###CHR### placeholder in VEP custom annotation to allow annotating from multiple file. But as web VEP job submission happen via eHive and eHive has a logic to param replacement with anything #TOKEN# it is creating failure.

The solution tried here is to escape # in the ###CHR### in the custom config with double backlash and replace them back again before submitting the VEP job.

Need for using double backlash -
- The config JSON is read by using Perl JSON module and it cannot parse a single backlash in the config.
Need for replacing the backlash again
- VEP can support ###CHR### or an escaped version with single backlash \#\#\#CHR\#\#\# but perl JSON module converts \\ --> \\ (not single backlash). That's we need to replace them back again.

sandbox test url - http://wp-np2-35.ebi.ac.uk:7070/Homo_sapiens/Tools/VEP/Results?tl=nqITqTx9D5mEJyI8-17

**Paralogue plugin**:
Update Paralogue plugin file name to match the latest data.

### DIFF
--- a/tools/conf/vep_custom_web_config.json
+++ b/tools/conf/vep_custom_web_config.json
@@ -7,7 +7,7 @@
       "description": "Report allele frequencies from the NIH AllOfUs study",
       "section": "Variants and frequency data",
       "params": {
-            "file": "/homo_sapiens/GRCh38/variation_genotype/AllOfUs/AllOfUs_chr###CHR###.vcf.gz",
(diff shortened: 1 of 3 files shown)


jamie-m-a (approved):
Does the trick - thanks Nakib

jamie-m-a (approved):
Thanks for updating paralogs as well.

PR #867, opened 2025-04-10
Fix some broken hrefs

EbiArnie:


azangru (approved):
Reminder to ourselves: the reason this PR is against main rather than postreleasefix/114 is because the white space at the start of a link only presents problems to our crawler script that runs during the release process (to create search xmls for help&docs); and therefore this will only become relevant again when we prepare for release 115. Real browsers don't care about such trifles.

veidenberg (approved):

PR #868, opened 2025-04-20
AllOfUs webVEP: remove fields and disable GRCh37 option

nakib103:
AllOfUs webVEP option
- Remove max_subpop option
- Disable option for GRCh37 website

Generic custom config improvement
- Remove identifier field that contains location which is duplicate info 

sandbox url - http://wp-np2-35.ebi.ac.uk:7070/Homo_sapiens/Tools/VEP/Results?tl=t1raqNqIYFGZdjhJ-36


sarahhunt:
Thanks for the fixes @nakib103 - this looksa good.

PR #869, opened 2025-04-22
Version++

sgiorgetti:


TamaraNaboulsi (approved):
Added/updated ini files, pictures, and SiteDefs for new and updated genomes for vertebrates. Matches list of removed, added, and updated vertebrate genomes. Looks good.

PR #870, opened 2025-04-28
Add update MaveDB data information on VEP What's new - 114

nakib103:
sandbox - http://wp-np2-35.ebi.ac.uk:7070/info/docs/tools/vep/script/vep_download.html#new


jamie-m-a (approved):
LGTM

PR #873, opened 2025-05-12
Shuffle sheep variation example to match reference change - 115

nakib103:
`ovis_aries` is now rambouillet instead of `ovis_aries_rambouillet`
similarly, `ovis_aries_texel` is now texel instead of `ovis_aries`

Moved around their data in INI files accordingly.


jamie-m-a (approved):
Thanks Nakib, I hadn't thought to check ini files!

PR #875, opened 2025-05-28
Add back max subpop column for AllOfUs in web VEP - 115

nakib103:
We had to remove the `max_subpop` field for AllOfUs option in web VEP because they were missing from the data in this PR - https://github.com/Ensembl/public-plugins/pull/868

With the addition of this data we are adding back the field.

sandbox url for test - 
http://wp-np2-35.ebi.ac.uk:7070/Homo_sapiens/Tools/VEP/Results?tl=nStkHAdk33R9mNrs-62


jamie-m-a (approved):
LGTM!

PR #881, opened 2025-06-17
Updating the number of Mus musculus strains

Simarpreet-Kaur-Bhurji:
I have updated the total number of strains in the Mus_musculus_strains.inc document, in light of the Murinae EPO update for release 115.


sgiorgetti (approved):
LGTM

PR #884, opened 2025-07-05
web VEP: add overlap cutoff to SV custom option (e115)

nakib103:
Add overlap cutoff options for gnomAD and ClinVar Significance SV options.
When selected user can chose if they want 80%, 90% , 100% overlap or an exact match. 

See input form - [sandbox url](http://wp-np2-35.ebi.ac.uk:7070/Homo_sapiens/Tools/VEP?tl=SgSjoiHiQVM4rRHt-91)

Also 
- removed clinical_source, REGIONID and ORIGIN from ClinVar output.
- removed ClinVar somatic as in the website they are appearing as germline.
- fix in filter_vep to use `-no_missing_consequence_warning` to avoid random outputs (related PR - https://github.com/Ensembl/ensembl-vep/pull/1900). 

sandbox test url - http://wp-np2-35.ebi.ac.uk:7070/Homo_sapiens/Tools/VEP/Results?tl=ZS2u5uQFsMtuUtlp-121

### DIFF
--- a/tools/modules/EnsEMBL/Web/TmpFile/VcfTabix.pm
+++ b/tools/modules/EnsEMBL/Web/TmpFile/VcfTabix.pm
@@ -116,7 +116,7 @@ sub content_iterate {
     }
 
     # build the commandline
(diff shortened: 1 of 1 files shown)


likhitha-surapaneni (approved):
Output as expected, thanks @nakib103

PR #886, opened 2025-07-09
Update Vertebrates "Protein trees" and "Homology types" pages

ens-sb:
This PR updates the "Protein trees" and "Homology types" pages:

- Emphasizes in "Protein trees" that between-species paralogies are not stored.
- Correct the duplications score thresholding in "Homology types".
- Applies small language fixes throughut.

Related ticket: ENSCOMPARASW-8498

### DIFF
--- a/ensembl/htdocs/info/genome/compara/homology_types.html
+++ b/ensembl/htdocs/info/genome/compara/homology_types.html
@@ -29,14 +29,14 @@
 <li>between-species paralogues – only as exceptions</li>
 </ul>
 
-<p>Genes in different species and related by a speciation event are defined as <strong>orthologues</strong>. Depending on the number of genes found in each species, we differentiate among 1:1, 1:many and many:many relationships. Please, refer to the figure where there are examples of the three kinds.</p>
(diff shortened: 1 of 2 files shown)


ensembl/htdocs/info/genome/compara/homology_types.html
```diff
@@ -60,34 +60,43 @@ <h2 id="paralogues">Paralogues</h2>
...
+<p>Right after building the multiple alignment, we detect pairs of genes that lie close to each other (< 1 MB) in the same sequence region and the same strand (see GeneA1/GeneA2 below). They are forced to be paired in the gene trees under gene_split events, and represent our most confident set. When there is no (or little) overlap between the gene fragments in the same species and they lie in different sequence regions in the assembly (see GeneB1/GeneB2 in the image below), we let TreeBeST <a href="#fn1">[1]</a> decide their best position in the tree. They are linked by duplication nodes, and have a lower confidence index. </p>
```
twalsh-ebi, line 83:
```suggestion
<p>Right after building the multiple alignment, we detect pairs of genes that lie close to each other (&lt; 1 MB) in the same sequence region and the same strand (see GeneA1/GeneA2 below). They are forced to be paired in the gene trees under gene_split events, and represent our most confident set. When there is no (or little) overlap between the gene fragments in the same species and they lie in different sequence regions in the assembly (see GeneB1/GeneB2 in the image below), we let TreeBeST <a href="#fn1">[1]</a> decide their best position in the tree. They are linked by duplication nodes, and have a lower confidence index. </p>
```

PR #891, opened 2025-07-15
variatiion_stats_updates

jamie-m-a:
Doc / stats updates @nakib103 can you double check this, it was all generated by the script but some look wrong to me. Probably need a sandbox


nakib103:
@jamie-m-a,

Besides above, we are also missing updates to the following pages -
http://wp-np2-35.ebi.ac.uk:7070/info/genome/variation/species/sources_documentation.html
http://wp-np2-35.ebi.ac.uk:7070/info/genome/variation/species/species_detailed_counts.html

jamie-m-a:
Apologies I was pointing at the wrong version of the pp repo when running. I'm going to close this and open a new one in a little while, will be too messy to reconcile.

PR #892, opened 2025-07-15
vep_plugins_updates

jamie-m-a:


nakib103 (changes requested):
@jamie-m-a,

We are missing the new [MechPredict](https://github.com/Ensembl/VEP_plugins/pull/772) plugin in the list,

Everything else look good.

PR #908, opened 2025-08-07
EPMC links

jamie-m-a:
Adding some additional citation links for VEP


nakib103 (approved):
LOTM!~

jamie-m-a:
Not Nakib, but I think the ~ is just flair, not a comment on the PR quality :)

PR #910, opened 2025-08-18
Update pathogenicity docs (116)

nakib103:
http://wp-np2-35.ebi.ac.uk:7070/info/genome/variation/prediction/protein_function.html


jamie-m-a (changes requested):
Minor text change in the comment and can we also change the screenshot:
http://wp-np2-35.ebi.ac.uk:7070/info/genome/variation/prediction/pathogenicity_predictions.png

To one takenm on the updated code, so it doesn't show mutationassessor

jamie-m-a (approved):
Thanks for the updates Nakib!

PR #911, opened 2025-08-19
Improve account registration process

veidenberg:
This PR patches few issues in our user account registration process, which is a common source of RT tickets.
* Provide the user an option to resend the account activation email
* Add server-side check for non-Latin characters in the account registration form


shashankbrgowda:
Its working as expected. Thanks @veidenberg

PR #915, opened 2025-09-24
Fix sheep configs for population displays

likhitha-surapaneni:
Population displays for sheep are not working as expected due to issues in config. This PR updates the configs to enable displays as expected.

Sandbox:
Ovis_aries: http://wp-np2-35.ebi.ac.uk:6080/Ovis_aries/Variation/Population?db=core;r=1:164280-165280;v=rs1090132668;vdb=variation;vf=38803995
Ovis_aries_texel: http://wp-np2-35.ebi.ac.uk:6080/Ovis_aries_texel/Variation/Population?db=core;r=1:1478138-1479046;v=rs1090132668;vdb=variation;vf=58541878

### DIFF
--- a/ensembl/conf/json/ovis_aries_vcf.json
+++ b/ensembl/conf/json/ovis_aries_vcf.json
@@ -1,7 +1,7 @@
 {
   "collections": [
     {
-      "id": "nextgen_sheep_iroa",
+      "id": "nextgen_sheep_rambouillet_iroa",
       "species": "ovis_aries",
       "assembly": "ARS-UI_Ramb_v3.0",
       "type": "local",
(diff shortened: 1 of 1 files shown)


ensembl/conf/json/ovis_aries_vcf.json
```diff
@@ -3,9 +3,9 @@
     {
       "id": "nextgen_sheep_iroa",
```
nakib103, line 4: advisable to update the id to `nextgen_sheep_rambouillet_iroa`

same as [ensembl-variation](https://github.com/Ensembl/ensembl-variation/blob/b7c2637ef23cf37aebe199266088918949674d94/modules/Bio/EnsEMBL/Variation/DBSQL/vcf_config.json#L660)

The id should be unique, it might not make an issue if only this json is loaded but it is better to be safe.

nakib103 (changes requested):

nakib103 (approved):
LGTM!~ thanks @likhitha-surapaneni

PR #951, opened 2026-04-02
Updated VEP max_sv_size arg docs to explain new disable option

mluypaert:
New option implemented in https://github.com/Ensembl/ensembl-vep/pull/1996 (so dependent on its approval and merging)

### DIFF
--- a/docs/htdocs/info/docs/tools/vep/script/vep_download.html
+++ b/docs/htdocs/info/docs/tools/vep/script/vep_download.html
@@ -559,7 +559,7 @@ cd ensembl-vep-release-[[SPECIESDEFS::ENSEMBL_VERSION]]/</pre>
         <ul>
           <li><a href="vep_options.html#opt_clin_sig_allele">--clin_sig_allele</a>: report allele specific clinical significance.</li>
           <li><a href="vep_options.html#opt_mane">--mane</a>: report if a transcript is the MANE Select.</li>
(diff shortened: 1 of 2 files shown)


docs/htdocs/info/docs/tools/vep/script/vep_download.html
```diff
@@ -559,7 +559,7 @@ <h4>New in version 97 <i>(July 2019)</i></h4>
         <ul>
           <li><a href="vep_options.html#opt_clin_sig_allele">--clin_sig_allele</a>: report allele specific clinical significance.</li>
           <li><a href="vep_options.html#opt_mane">--mane</a>: report if a transcript is the MANE Select.</li>
-          <li><a href="vep_options.html#opt_max_sv_size">--max_sv_size</a>: extend the maximum Structural Variant size Ensembl VEP can process.</li>
+          <li><a href="vep_options.html#opt_max_sv_size">--max_sv_size</a>: Limit the maximum Structural Variant size Ensembl VEP will process to reduce potential memory usage.</li>
```
nakib103: It is probably not a good idea to update release note of e97. The statement could have been correct then.

nakib103 (changes requested):

docs/htdocs/info/docs/tools/vep/script/vep_download.html
mluypaert: Oops, good catch. I've reverted this change.

nakib103 (approved):

PR #952, opened 2026-04-07
More variation public doc updates

nakib103:
Some more updates to public docs for variation -
- Revert back protein function pages mentioning we are using 3 bits for prediction and add a note it is only for AlphaMissense and ESM1b.
- Update filter_vep docs mis-using `is` operator for SIFT/PolyPhen score when both prediction and score is present.
- Fix rows for rabbit and japanese quail in species_data_types.html


jamie-m-a (approved):
LGTM

PR #953, opened 2026-04-13
Reforge www glossary links

twalsh-ebi:
This PR would reforge glossary links and link contexts in the documentation of VEP and other topics.


likhitha-surapaneni (approved):
This looks good to me @twalsh-ebi , thank you!

PR #957, opened 2026-04-30
Enhancement/ajax calls

sgiorgetti:
## Summary

This PR reduces unnecessary masthead AJAX traffic by lazy-loading account dropdown content and only refreshing the tools Jobs tab when the browser is likely to have tools activity.

## Changes

- Lazily load `/Ajax/accounts_dropdown` when the account dropdown is first opened or focused, instead of requesting it on every page load.
- Render the stable account/login link server-side so the masthead remains usable before dropdown content is fetched.
- Add account dropdown loading/caching state, callback handling, namespaced event bindings, and forced refresh support for account modal updates.
- Limit the masthead bookmark preview to 5 entries while keeping full bookmark management available from the account page.
- Avoid eager `*/Ajax/tools_tab` requests unless there are recent/current tools jobs, a `tl` parameter, or a persisted browser hint.
- Store a lightweight `ensembl.hasToolsJobs` localStorage hint so later non-Tools pages can show Jobs without probing users who have never submitted jobs.
- Refresh the tools masthead after ticket submission and after tools save/delete refresh responses.

## Testing

Manual testing recommended:

- Confirm logged-out users see the Login/Register link without an initial accounts dropdown AJAX request.
- Confirm logged-in users see the account link immediately and dropdown content loads on first open/focus.
- Confirm account dropdown links still work for bookmarks, account details, manage bookmarks, and logout.
- Confirm submitting a tools job updates the Jobs tab and persists the tools-job hint.
- Confirm deleting/saving tools jobs refreshes the masthead state correctly.
- Confirm users with no tools jobs do not trigger unnecessary tools-tab AJAX calls on unrelated pages.

### DIFF
--- a/tools/htdocs/components/10_Masthead_Tools.js
+++ b/tools/htdocs/components/10_Masthead_Tools.js
@@ -19,6 +19,14 @@
 
 Ensembl.Panel.Masthead = Ensembl.Panel.Masthead.extend({
 
+  toolsJobStateKey: 'ensembl.hasToolsJobs',
+
+  constructor: function (id) {
+    this.base(id);
+
+    Ensembl.EventManager.register('toolsRefreshMasthead', this, this.refreshToolsTab);
+  },
+
   init: function () {
     this.base();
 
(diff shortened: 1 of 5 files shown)


azangru:
I have tried to review the changes to `10_Masthead_Tools.js`; and the biggest question that I have about it is: does this file even have a purpose anymore?

Presumably, it does something to a tools dropdown (see [CSS selectors here](https://github.com/Ensembl/public-plugins/blob/6956bae45b3388b0f1eafa9d92b831fdd64b73dd/tools/htdocs/components/10_Masthead_Tools.js#L33-L34)). I have never seen the tools dropdown. Apparently, according to CSS selectors, it is this element:

image

But the dropdown doesn't open. If I try to open it, I am getting an error:

image

On the plants site, there isn't even a visual indicator about the presence of the dropdown:

image

I think, given that it is broken, we can just delete the masthead tools js file entirely. That should solve the problem with the ajax.

PR #967, opened 2026-06-05
Variation docs - add the VEP PDF docs

nakib103:
sandbox:
http://wp-np2-35.ebi.ac.uk:7070/info/docs/tools/vep/online/VEP_web_documentation.pdf
http://wp-np2-35.ebi.ac.uk:7070/info/docs/tools/vep/script/VEP_script_documentation.pdf

PR #969, opened 2026-06-08
Variation docs - add the VEP PDF docs

nakib103:
same as https://github.com/Ensembl/public-plugins/pull/967 and but to `postrealeasefix/116` branch


jamie-m-a (approved):
Thanks rebasing, looks good!

PR #972, opened 2026-06-24
Replace larger images with placeholder in species tree view

twalsh-ebi:
This PR would replace larger images with a placeholder in the species tree view, ensuring memory usage would be kept within reasonable limits.

Example: [sandbox](http://wp-np2-35.ebi.ac.uk:5092/info/about/speciestree.html) vs [live site](https://www.ensembl.org/info/about/speciestree.html)

### DIFF
--- a/widgets/modules/EnsEMBL/Web/Document/HTML/Compara/SpeciesTree.pm
+++ b/widgets/modules/EnsEMBL/Web/Document/HTML/Compara/SpeciesTree.pm
@@ -88,6 +88,7 @@ sub render {
   $tree_details->{'default_tree'} = $mlss->has_tag('default_tree') ? $mlss->get_value_for_tag('default_tree') : (keys %{$tree_details->{'trees'}})[0];
  
   # Go through all the nodes of all the trees and get the tooltip info
+  my $default_image;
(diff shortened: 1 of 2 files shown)


widgets/modules/EnsEMBL/Web/Document/HTML/Compara/SpeciesTree.pm
```diff
@@ -115,6 +116,21 @@ sub render {
           if ($content) {
             $sp->{icon} = 'data:image/png;base64,'.encode_base64($content);
           }
+
+          if (length($sp->{icon}) > 1_000_000) {  # We assume one million characters is sufficient for a species-tree species image.
```
sgiorgetti: As agreed the limit should be defined and pulled from SiteDefs

twalsh-ebi: Thanks for this suggestion @sgiorgetti and @TamaraNaboulsi. As discussed, I've made this threshold configurable as `BASE64_MAX_SIZE`.

sgiorgetti (approved):
LGTM

PR #978, opened 2026-07-31
Update kent build information for new GNU versions

nakib103:
A user has reported (see https://github.com/Ensembl/ensembl-vep/issues/1412) that kent source building fails for VEP for latest ubuntu version and pointed out setting the compiler to use GNU 17 solves the issue.

We can safely set `-std=gnu17` as GCC has support for older GNU tools.

sandbox view - http://wp-np2-35.ebi.ac.uk:7070/info/docs/tools/vep/script/vep_download.html#bigfile


richysix (approved):
Looks good.

